PR #1, opened 2017-07-24
docs(readme): fix grammar

dferber90:
> The HoC to conditionally render a component based on a feature toggle's state. It accepts the feature toggle name and ~~a~~ an optional component to be rendered in case the feature is disabled.


tdeekens:
🙏

PR #6, opened 2017-08-05
Adds `shouldInitialize` prop to `ConfigureFlopflip`

tdeekens:
- For both packages react-redux and react-broadcast
- Extracts flag subscription into component in react package
   - Couples logic to `listen` in componentDidMount and componentWillReceiveProps
- Fixes up callback props in `launchdarkly-wrapper` to have better naming


tdeekens:
This is what we talked about briefly @emmenko

PR #14, opened 2017-08-06
✨ Add `injectFeatureToggle` HoC

### LINKED ISSUE #10: Add `injectFeatureToggle` HoC
> Injecting a single feature toggle is currently "only" supported via the `withFeatureToggle' HoC.

Whenever a component should be toggled the `withFeatureToggle` HoC can be used. Whenever more sophisticated decisions need to be made upon multiple flags the `injectFeatureFlags` HoC can also be used.

However, in some cases not a whole component needs to be toggled as per `withFeatureToggle`. When injecting only one toggle `injectFeatureToggles` can become a bit verbose as you have to specify an array and in the component have to pluck the toggle off the `featureToggles`-prop again.

```js
injectFeatureToggle('isButtonOn')(Header)
```

Would just inject `isFeatureOn` into the `Header` as a `prop` based on the state of `isButtonOn`. 

An addition would be to specify the `propName` for the flag which may default to `isFeatureOn` so the API becomes

```js
injectFeatureToggle('isButtonOn', 'shouldShowYellowButton')(Header)
```

tdeekens:
Adds a HoC to inject a single feature toggle.

- Refactors with-feature-toggle to use the component
- Renames requestedFeatureToggles
- Fixes tests of branchUntoggled while adjusting it to accept a flagName

Closes #10

### DIFF
--- a/demo/src/App.js
+++ b/demo/src/App.js
@@ -5,7 +5,7 @@ import classNames from 'classnames';
 import {
   ConfigureFlopFlip,
   withFeatureToggle,
-  injectFeatureToggles,
+  injectFeatureToggle,
   FeatureToggled,
 } from '@flopflip/react-redux';
 // change to `from '@flopflip/react-broadcast'` and everything will just work wtihout redux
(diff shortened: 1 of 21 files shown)


tdeekens:
Should be ready to go. Letting this sink in for later review.

PR #22, opened 2017-08-08
✨ Add automatic changing user context when already initialized

### LINKED ISSUE #11: Add ability to change the user context
> LaunchDarkly's js-sdk allow switching the user context which could be ported to flopflip too.

The js-sdk has the API method `identify` to change a users' context. An example would be to have an anonymous user on a login page but an authed within the application. This would flush initial flags down to components and later new ones based on the authed user.

Currently flopflip supports the `shouldInitialize` callback prop to defer initialization with LaunchDarkly on e.g. a login page. Until then the static `defaultFlags` would be flushed to components. Changing the users' context would allow even more flexibility towards that.

Things to watch out for

- Empty all flags when user context changes to not mix old with new
- Keep client on component instance in `FlagsSubscription` to invoke `identify`
- Just watching to changing `key` in `componentWillReceiveProps` should be enough

tdeekens:
Closes #11 

Please refer to issue for feature desciption.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ node_js:
   - "6"
   - "8"
 cache: yarn
-script: yarn test:coverage:ci
+script: yarn test:ci
 notifications:
    email: false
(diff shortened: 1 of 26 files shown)


tdeekens:
Not fully done. Needs testing if flag values do not mix.

tdeekens:
Voila @dferber90

dferber90:
🌈

PR #26, opened 2017-08-27
Improve multi variante support for toggles

tdeekens:
> Multivariate toggles may become an important part of people using this library but our integration with them can be improved.

Currently multivariate toggles are only supported when using the `injectFeatureToggle` HoC by then "switching" over the injected flag value.

https://github.com/tdeekens/flopflip/blob/master/demo/src/App.js#L33-L49

This pull request eases working with multivariates for `withFeatureToggle` and `FeatureToggled`.

```jsx
// Without any variate we would default to `true`

withFeatureToggle({ flagName: 'fooFlag' })(ComponentToBeToggled)

// we could also add a property to become 

withFeatureToggle({ flagName: 'fooFlag', flagVariate: 'fooVariate' })(ComponentToBeToggled)
```

For the `FeatureToggled` component this would mean props would also change

```jsx
<FeatureToggled flagName="fooFlag" flagVariate="fooVariate">
     <Feature />
</FeatureToggled>
```

In this scenario both components only operate on one variate of a flag each. The assumption is that an app would have features structured as in

```txt
- /components
   - feature-variate-1
   - feature-variate-2
```

an each would wrap a variate using `withFeatureTogggle`. Using `FeatureToggled` would require n-components

```jsx
<FeatureToggled flagName="fooFlag" flagVariate="fooVariate1">
     <Feature />
</FeatureToggled>

<FeatureToggled flagName="fooFlag" flagVariate="fooVariate2">
     <Feature />
</FeatureToggled>
```

:shipit: ing this would entail a breaking change to the API. `withFeatureToggle` would accept an `Object` as its first argument over a `String`. Whereas, `FeatureToggled` would receive `flagName` and `flagVariate` as props.


tdeekens:
We would obviously remove the `flag*` from any prop or argument and have APIs such as `withFeatureToggle({flag: 'foo', variate: 'bar})` and `FeatureToggled flag="foo" variate="bar"`.

Curious on thoughts of @emmenko and @dferber90.

dferber90:
Without an understanding of the existing APIs, some general thoughts:

There's probably two ways in which variates are used:
1. Variates of a component are all in the same place and using a different variate only needs to render a different component. An example would be testing multiple versions of a datepicker.
1. A page where the layout differs depending on the variate. So one variate may not be rendered in the exact place of the other variate. An example could be testing different locations for a call-to-action button.

Both cases probably require slightly different APIs. This issue seems to tackle the first use case.

tdeekens:
Yes, if one it tackle the first issue. The latter would probably we a case for `injectFeatureToggle`. Both a `Footer` and `Header` would get a toggle injected with the variate attached and can render accordingly - hopefully disjoint. The first approach "problem" would probably be solved by `withFeatureToggle`. You might have two date-pickers (internally wrapping) and implementation of say a "Yahoo" and "Airbnb" date picker. Their default export would then be `withFeatureToggle('datepicker-relaunch', 'airbnb'` and `withFeatureToggle('datepicker-relaunch', 'yahoo'`.

emmenko:
I'm not sure I fully understand the use case here, but it seems fine to me.

> breaking change to the API

We can simply be backwards compatible by checking the `typeof` the argument. Obviously if you pass an object you can use this `variate` option. You can also log a warning if the "deprecated" API is used.

> We would obviously remove the `flag*`

👍

tdeekens:
> checking the typeof and console.warn

Is a nice idea. Will probably implement that.

> In conjunction with removing the flag prefixing this seems like a good idea to minimise the impact of this change.

Assuming that this feature will actually ever be needed.

tdeekens:
The safe extraction with a deprecation warning when not passing an object (until 3.0.0) is part of 
 aed22f0.

PR #31, opened 2017-09-03
Feature: Default variate values on component level

tdeekens:
> This pull request adds the feature of default variate values on component level.

Even though `flopflip` offers `defaultValues` during configuration through `ConfigureFlopflip` there are uses cases for default values on component level. These components are `FeatureToggled` and `withFeatureToggle`.

Given a feature toggle is not defined (does not exist) this allows to opt-in to turn the feature on. Whereas before it was then turned off.

```js
// Would otherwise default to `false` (without `defaultFlags` on `ConfigureFlopflip`)
withFeatureToggle({ flag: 'signup-v2', defaultVariateValue: true })
// this reads strange but means that it just default to the `signup-modal` variate if toggle doesn't exist
withFeatureToggle({ flag: 'signup-v2', variate: 'signup-default', defaultVariateValue: 'signup-modal' 
// this would make more sense
withFeatureToggle({ flag: 'signup-v2', variate: 'signup-modern', defaultVariateValue: 'signup-default'})

```

```jsx
// again otherwise this would default to `false`
<FeatureToggled flag="signup-v2" defaultVariateValue={true}>
...
</FeatureToggled>

// same as abvoe
<FeatureToggled flag="signup-v2" variate="signup-modal" defaultVariateValue="signup-modal">
...
</FeatureToggled>
// same as abvoe
<FeatureToggled flag="signup-v2" variate="signup-modern" defaultVariateValue="signup-default">
...
</FeatureToggled>
```

The naming of `defaultVariateValue` and `defaultVariate` is deliberate as the it needs to integrate with multivariate and toggles. Moreover it should be consistent across all components.


tdeekens:
Feature will be removed but other improvements from this branch will be kept.

Issue #40, opened 2017-09-16
[Idea]: Decouple from LaunchDarkly (add more adapters)


tdeekens:
> Currently `flopflip` is coupled to "only" LaunchDarkly as a provider for toggles this can change.

The library could potentially allow multiple providers for toggles and was actually built from the ground up to do so. Another `packages/*` to integrate with another provider and the configuration on of a consumer facing package should then pick any of the integrations behind the scenes. The API for an integration has a quite minimal surface area:

1. Initialize (with allowing to defer it)
2. Flush flags
3. Flush updates

An integration is in charge of "doing" the targeting etc.

A simple idea is to integrate with an autoupdating/polling/streaming localstorage integration. Any app could, when receiving flags from it's backend, flush these to localstorage where the adapter/integration picks them up and they're automatically flushed to redux or the broadcasting variant.

tdeekens:
1. This would probably involve renaming `ld-wrapper` to `launchdarkly-adapter`
2. This would include adding a `localstorage-adapter`
3. This would include having to pass in an adapter to `ConfigureFlopFlip` manually

Possible examples to configure adapters

```jsx
import launchdarklyAdapter from '@flopflip/launchdarkly-adapter';

<ConfigureFlopFlip adapter={launchdarklyAdapter} adapterArgs={{ clientSideId: '123' }}>
  <App />
<ConfigureFlopFlip>

//..

import localstorageAdapter from '@flopflip/localstorage-adapter';

<ConfigureFlopFlip adapter={localstorageAdapter}>
  <App />
<ConfigureFlopFlip>
```

or something like

```jsx
import launchdarklyAdapter from '@flopflip/launchdarkly-adapter';

launchdarklyAdapter.configure({ ldClientSideId: '123' })

<ConfigureFlopFlip adapter={launchdarklyAdapter}>
  <App />
<ConfigureFlopFlip>
```

tdeekens:
The localstorage adapter would 

1. Define a "constant" as a key to where flags are status data is persisted
2. Expose the API above (same for the launchdarkly-adapter`
3. Poll localstorage and "emit" changes though the `onFlagsStateChange` callback

tdeekens:
Another adapter could be an `url-adapter` which reads the flag statuses from the url. Behind the scenes it could use something like https://github.com/ReactTraining/history

tdeekens:
Another dapter could be an `api-adapter` which behind the scenes fetches flags from some service.

```js
[...]

PR #48, opened 2017-10-16
Refactor `flopflip` and setup to support notion of adapters

tdeekens:
This pull request relates to #40 and implements the notion of adapters.

During configuration `ConfigureFlopFlip` must receive an `adapter` and `adapterArgs`. The `adapter` must have the methods:

1. `configure`: which will be called when `shouldConfigure` is `true` during `componentDidMount` or `componentDidUpdate` being passed the `adapterArgs`
2. `reconfigure`: which will be called whenever the `adapter` reports as `isConfigured` during `componentDidUpdate`
3. `isConfigured`: indicates if the adapter has been successfully configured
4. `isReady`: indicates if the adapter is ready (when integrating with async external services)

The first `launchdarkly-adapter` for instance allows the switch the user context when `shouldReconfigure` on `ConfigureFlopFlip` is `true`. This is implemented in it's `reconfigure` method. A `localstorage-adapter` should e.g. wipe localstorage or switch to another storage slice.

The `adapterArgs` must be of the following form:

1. `user`: any user data needed to identify a user
2. `onFlagStateChange`: triggered by an adapter whenever any flag status changes 
3. `onFlagStatusChange`: triggered by an adapter whenever the adapter becomes ready or reports other status

In case of the `launchdarkly-adapter` an additional `adapterArg` is the `clientSideId` which will be passed along to initiate communication with LaunchDarkly.


tdeekens:
Next `react-broadcast` and `react-redux` have to forward and integrate.

tdeekens:
Alright packages are integrated with adapters, readme is updated. This needs some fresh 👁  some time.

tdeekens:
Essentially this is the setup change

```js
import { ConfigureFlopFlip } from '@flopflip/react-redux';
import adapter from '@flopflip/launchdarkly-adapter';

<ConfigureFlopFlip adapter={adapter} adapterArgs={{clientSideId, user}}>
  <App />
</ConfigureFlopFlip>;
```

tdeekens:
Alright little update the the `adapter` API. We dropped the need for adapters to expose `isReady` and `isConfigured` functions. This state now rightfully lives within `FlagsSubscription` which sets up the `adapter` passed. With this an `adapter` only needs to expose a `configure` and `reconfigure` function. Both should return a `Promise` however. In the case of LaunchDarkly initialisation is async and for others this might also be the case (WebSockets, HTTP etc). The `FlagsSubscription` component can nicely hook into those async functions and maintain the adapter state accordingly to e.g. invoke `reconfigure` after initial configuration. 

Furthermore, as per suggestion by @dferber90 the `ConfigureFlopflip` component now does not receive `shouldConfigure` and `shouldReconfigure` but only one `shouldDeferAdapterConfiguration` which defaults to `false` and can be flipped by a SPA until e.g. it has loaded user data.

tdeekens:
Arr, darn you coverage.

PR #65, opened 2017-11-20
Add splitio adapter

Kerumen:
This add the adapter for the split.io service (https://www.split.io/).

I adapted it from the LaunchDarkly adapter with few differences:
* There is no boolean flags so I converted `on` and `off` to boolean and let others pass through
* 2 calls have to be made to get first, the flags, then, the values
* There is no concept of `user` because split.io uses segments, but you can pass a map of attributes in the `getTreatment[s]` calls so I adapted this concept with the notion of `user` the flipflop's API has
* The listener doesn't returns changed flags. It only indicates a change. So flags must be refetched.

Finally, the split.io SDK returns a `factory` where you can get either the `.client()` or the `.manager()`. I had trouble to test this because I'm pretty new to Jest so I let 2 tests failed (and one implementation commented out). If you can help me cover this, it would be very appreciated!
Otherwise, removing the tests would work too :D

### DIFF
--- a/packages/splitio-adapter/modules/adapter/adapter.js
+++ b/packages/splitio-adapter/modules/adapter/adapter.js
@@ -56,7 +56,7 @@ const ensureUser = user => ({
 const initializeClient = (authorizationKey, user) => {
   const factory = splitio({
     core: {
-      authorizationKey: authorizationKey,
+      authorizationKey,
       key: user.key,
     },
   });
--- a/packages/splitio-adapter/modules/adapter/adapter.js
+++ b/packages/splitio-adapter/modules/adapter/adapter.js
@@ -53,7 +53,7 @@ const ensureUser = user => ({
   ...user,
 });
 
-const initializeUserContext = (clientSideId, user) => {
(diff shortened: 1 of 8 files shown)


tdeekens:
Wow, thanks a lot for putting up this PR and sorry for the bit delayed response. I will take the time to go over things this weekend the latest.

Kerumen:
@tdeekens I did merge / rebase but I can't get the conflicts on my computer. It seems all clear on the command line but GitHub detect a conflict. It's weird..

tdeekens:
Hm, strange. Can you maybe repoint this PR to a new branch on my repo I just created (just click the edit icon)? It's called `feature/splitio-adapter`. Then we can merge it into that branch on I can take over from there as I have push access etc which is also good for fixing the build etc.

Kerumen:
@tdeekens Done! I resolved the `yarn.lock` conflicts :)

tdeekens:
Very odd. Thanks. I will try to get this in and will ping our one last time before the initial release if that's fine by you.

PR #67, opened 2017-11-23
Add `splitio-adapter` package

tdeekens:
Thanks for @Kerumen we will soon have an adapter for splitio. This branch aims to fix the last issues with build setup, documentation etc.

### DIFF
--- a/packages/splitio-adapter/modules/adapter/adapter.js
+++ b/packages/splitio-adapter/modules/adapter/adapter.js
@@ -56,7 +56,7 @@ const ensureUser = user => ({
 const initializeClient = (authorizationKey, user) => {
   const factory = splitio({
     core: {
-      authorizationKey: authorizationKey,
+      authorizationKey,
       key: user.key,
     },
   });
--- a/packages/splitio-adapter/modules/adapter/adapter.js
+++ b/packages/splitio-adapter/modules/adapter/adapter.js
@@ -53,7 +53,7 @@ const ensureUser = user => ({
   ...user,
 });
 
-const initializeUserContext = (clientSideId, user) => {
+const initializeClient = (clientSideId, user) => {
   const factory = splitio({
     core: {
       authorizationKey: clientSideId,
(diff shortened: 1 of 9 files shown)


tdeekens:
This is the build error I was expecting. I will dive into another round of `rollup` debugging another day.

tdeekens:
Couldn't resist. Debugging the plugin chain of `rollup` this comes from the `babel-plugin` which receives something like

```js
import '../../modules/es6.string.iterator';
import '../../modules/es6.array.from';
import '../../modules/_core';
import 'commonjs-proxy:../../modules/es6.string.iterator';
import 'commonjs-proxy:../../modules/es6.array.from';
import require$$2 from 'commonjs-proxy:../../modules/_core';
```

which looks like the `commonjs-plugin` processes the `core-js` dependencies which it shouldn't have to and shouldn't due to the `exclude` option.

tdeekens:
I literally am baffled. There is a `modulesOnly: true` option for the `plugin-resolve` in rollup which does some funky things. Which stems from [this](https://github.com/rollup/rollup-plugin-node-resolve/issues/91) issue and is exactly what we ran into. I am not even fully sure what it implies after reading the description five times. It seems like imports from a non ES module to an ES module without a `pkg.main` cause trouble at times to be detected correctly.

tdeekens:
I will rebase one last time removing the replacing of the version from the package but rather reading it using the `json-plugin`.

tdeekens:
Is there anything you see missing @Kerumen? Otherwise I'd be pretty confident to release this as `1.0.0` unless you want to test the code for instance again.

Kerumen:
@tdeekens No I'm fine, go release this! Thanks to give your time to investigate. The library just got better!

tdeekens:
Sure. Let me know if you see any other areas or room for improvement. Always eager to hear other people's perspectives.

tdeekens:
Voila. Let me know if there are problems.

https://github.com/tdeekens/flopflip/releases/tag/%40flopflip%2Fsplitio-adapter%401.0.0

PR #71, opened 2017-11-24
Replace camelcase module by Lodash camelcase

Kerumen:
We just had the same build problem you had when bundling all the dependencies. The problem is that Sindre Sorhus doesn't love the browser and target node. So he doesn't compile his modules.

We need to use this for the web and we use create-react-app. CRA doesn't compile the `node_modules` (https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-build-fails-to-minify).

Hence, I switched to `lodash.camelcase` which behaves exactly the same. I let you revert the special rollup config you did if you want to.

### DIFF
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.js
@@ -1,5 +1,5 @@
 import { initialize } from 'ldclient-js';
-import camelCase from 'lodash.camelcase';
+import camelCase from 'just-camel-case';
 
 const adapterState = {
   isReady: false,
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.js
@@ -1,5 +1,5 @@
 import { initialize } from 'ldclient-js';
-import camelCase from 'camelcase';
+import camelCase from 'lodash.camelcase';
 
 const adapterState = {
   isReady: false,
--- a/packages/launchdarkly-adapter/package.json
+++ b/packages/launchdarkly-adapter/package.json
@@ -34,8 +34,8 @@
     "enzyme-to-json": "^3.2.2"
   },
   "dependencies": {
-    "camelcase": "^4.1.0",
-    "ldclient-js": "^1.1.12"
+    "ldclient-js": "^1.1.12",
+    "lodash.camelcase": "^4.3.0"
   },
   "keywords": [
     "feature-flags",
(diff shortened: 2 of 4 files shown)


tdeekens:
Thanks. A bit strange as I would have thought the rollup build would transpile these commonjs modules over to esm. Maybe I‘d rather use [just.camelcase](https://github.com/angus-c/just/blob/master/README.md#just-camel-case) if possible. I had annoyances in the past with lodash bundles. Would that work for you? Thanks again!

tdeekens:
Moreover, we use the just libs for one or two other things abd I‘d rather stick to one land of utiliy fns. Moreover, lodash often comes with quite a bit of clutter due to their modular but entangled setup. Anyway, if just-camelcase doesn‘t work for you I am obviously totally fine with lodash.

Kerumen:
Yes, `just-camel-case` would be totally fine.
You already used lodash [here](https://github.com/tdeekens/flopflip/blob/a2ff89e00b53e38d015008e3d3c1e23eeda885e3/packages/react/package.json#L50) so i thought it would be fine. I make the change.

tdeekens:
Hah. You caught me there. just doesnt have any alternative. Id like too remove isNil at some point. Maybe before the release of this.

tdeekens:
Turned on builds for forks on CircleCI from now on. Checked this locally and it builds fine (just to make sure). Will merge. Thanks.

tdeekens:
Moep. just-camelcase didn't work as I expected and broke tests on `master`. Looking into it.

tdeekens:
At least builds will now be triggered on forks giving us a bit more safety.

tdeekens:
Not so nice. Sorry for hooking you to use just-camelcase cause this case won't work for the library `  camelCase('theQuickBrownFox'); // 'thequickbrownfox'` (from the docs).

tdeekens:
Voila, enjoy. Hope it resolves your "issues". Have a good weekend.

https://github.com/tdeekens/flopflip/releases/tag/%40flopflip%2Fsplitio-adapter%401.0.1

Kerumen:
Thanks! Will try this out on monday :)

angus-c:
@tdeekens @kerumen FWIW just-camel-case@1.0.0 should work the way you want. I actually updated to the new behavior because I got feedback that folks preferred that way. There's a moral there somewhere :) 

See: https://github.com/angus-c/just/commit/3195deb733272429b92548feb89c82f3ca4a4815#diff-4f09948ac318a1e5125ed279dc5fad60

tdeekens:
Thanks for the feedback Angus. Will look into it.

PR #74, opened 2017-11-28
Pass args to configure the splitio client

Kerumen:
The splitio JS client can be configured with a lot of options (https://docs.split.io/docs/javascript-sdk-overview#section-configuration).

This brings the ability to pass an `adapterArgs` prop with the options we want to configure.

```
<ConfigureFlopFlip
  adapter={adapter}
  adapterArgs={{ authorizationKey, user, core: { labelsEnabled: false } }}
>
```

If you want to nest this behind an object key (like `adapterConfig` or something), let me know!

### DIFF
--- a/packages/splitio-adapter/modules/adapter/adapter.js
+++ b/packages/splitio-adapter/modules/adapter/adapter.js
@@ -54,12 +54,13 @@ const ensureUser = user => ({
   ...user,
 });
 
-const initializeClient = (authorizationKey, user) => {
+const initializeClient = (authorizationKey, user, params) => {
   const factory = splitio({
     core: {
       authorizationKey,
       key: user.key,
     },
+    ...params,
   });
 
   return {
@@ -95,11 +96,13 @@ const configure = ({
   user,
   onFlagsStateChange,
   onStatusStateChange,
+  ...adapterArgs
 }) => {
   adapterState.user = ensureUser(user);
   const { client, manager } = initializeClient(
     authorizationKey,
-    adapterState.user
+    adapterState.user,
+    adapterArgs
   );
   adapterState.client = client;
   adapterState.manager = manager;
--- a/packages/splitio-adapter/modules/adapter/adapter.js
+++ b/packages/splitio-adapter/modules/adapter/adapter.js
@@ -54,13 +54,27 @@ const ensureUser = user => ({
   ...user,
 });
 
-const initializeClient = (authorizationKey, user, options) => {
+// NOTE: Little helper to omit properties from an object.
+// `lodash.omit` is too heavy in bundle size to add as a dependency.
+const omit = (obj, keys) =>
+  Object.entries(obj)
+    .filter(([key]) => !keys.includes(key))
+    .reduce(
+      (acc, [key, value]) => ({
+        ...acc,
+        [key]: value,
+      }),
+      {}
+    );
+
+const initializeClient = (authorizationKey, key, options = {}) => {
   const factory = splitio({
+    ...omit(options, 'core'),
     core: {
       authorizationKey,
-      key: user.key,
+      key,
+      ...options.core,
     },
-    ...options,
   });
 
   return {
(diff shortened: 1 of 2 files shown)


packages/splitio-adapter/modules/adapter/adapter.js
```diff
@@ -54,12 +54,13 @@ const ensureUser = user => ({
   ...user,
 });
 
-const initializeClient = (authorizationKey, user) => {
+const initializeClient = (authorizationKey, user, params) => {
```
tdeekens: Should this be named `options` in accordance with the adapter?

packages/splitio-adapter/modules/adapter/adapter.js
```diff
@@ -54,12 +54,13 @@ const ensureUser = user => ({
...
   const factory = splitio({
     core: {
       authorizationKey,
       key: user.key,
     },
+    ...params,
```
tdeekens: ^^

packages/splitio-adapter/modules/adapter/adapter.js
```diff
@@ -95,11 +96,13 @@ const configure = ({
   user,
   onFlagsStateChange,
   onStatusStateChange,
+  ...adapterArgs
```
tdeekens: Not sure about the spread. Should be just have an `options` key we expect to receive here? That way the adapter remains a bit more extensible and we don't "by accident" spread in too many things.

Kerumen: That's what I meant by nesting this in a key.
You'd prefer ano object like this: 
```
adapterArgs={{
  authorizationKey,
  user,
  options: {
    core: {
      labelsEnabled: false,
    },
  },
}}
```

tdeekens: Thanks for the feedback. Not each adapter will however have a key. Things like localstorage or custom http adapters. That‘s why I decided against it. For performance, no matter how deep an object or array it will we a new reference on render. At work I have a memoized function for creating the args which upon equal args returns and referentially equivilant object for the adapterArgs.

packages/splitio-adapter/modules/adapter/adapter.js
```diff
@@ -54,12 +54,13 @@ const ensureUser = user => ({
...
 });
 
-const initializeClient = (authorizationKey, user) => {
+const initializeClient = (authorizationKey, user, options) => {
   const factory = splitio({
     core: {
       authorizationKey,
       key: user.key,
     },
+    ...options,
```
tdeekens: Just a note. This might overwrite the above core propery when present again on the options as we do not take the adapterArga apart carefully further down.

Kerumen: True. Should we totally pass the `splitio()` config in the `adapterArgs` props?

tdeekens: Let me try to propose the solution Ive got in my head to validate it latest by the weekend if that‘s cool. I‘d prefer playing with, thinking out solutions before having to publish the adapter with breaking changes twice. Hope thats ok.

Kerumen: Yes no problem :)

tdeekens:
In case you've tested my suggestions from my PR towards your PR feel free to update this so we can move forward with this if needed.

Kerumen:
@tdeekens I left a comment 3 days ago on the PR :)

https://github.com/Kerumen/flopflip/pull/2#discussion_r153868478

tdeekens:
Ah alright. Sorry must have missed that. 🤗

tdeekens:
Sorry to bother you but: any news here?

Kerumen:
Hello! There is still the comment I left out on the PR :)

https://github.com/Kerumen/flopflip/pull/2#discussion_r153868478

tdeekens:
Was there anything missing here? The pending checks are just due to me renaming them on CircleCI. Otherwise I think we can merge and release this?

Issue #75, opened 2017-11-29
Rerender issues with `injectFeatureToggle`


Kerumen:
I noticed an issue when my component has `injectFeatureToggle`. It injects the selected feature flags in the component as a prop. However, this is an object. So each time the "update" action is fired, the component is rerendered.
In my app, I use split.io and I needed to inject some flags at the root. It leads to perf issues because my whole tree is rerendered every 15 seconds (splitio fire an update event every 15 secs in dev, 30 in prod, even if the flags have not changed).

I think, one way to solve this would be, in the `updateFlags` from `react-redux` and `react-broadcast`, to deep check if the flags have changed, and if not, avoid dispatching the action. This will fix this globally (for every adapters).
Or we do this specifically in the `splitio-adapter` (because he's the guilty one) and we avoid calling `onFlagsStateChange ` in the adapter.

For every cases, it would be better than dispatching an update action with an empty payload.

tdeekens:
`injectFeatureToggles` injects an object `injectFeatureToggle` should only inject a primitive value. But true, due to the immutability of flags and how we accept and handle flags from any adapter a re-render will happen. deep-equals might be something to consider. Maybe even shallow as flags should generally not be nested.

Kerumen:
I use `injectFeatureToggles` indeed, typo!

Flags are not nested so a shallow compare would easily do the job. What's the best location to implement this according to you?

tdeekens:
Hard to say. Yesterday react-broadcast (the lib) was released with a `compareValues` option. I'd like to have it in both world (redux and broadcast). 

1. Having it in adapters first feels okay-ish but I think we want it for all adapters after all and then it's repetitive code
2. We could also add it to [`injectFeatureToggles`](https://github.com/acdlite/recompose/blob/master/docs/API.md#onlyupdateforkeys) in `react` which is used by both `broadcast` and `redux` integrations
   - We could then e.g. just use https://github.com/acdlite/recompose/blob/master/docs/API.md#onlyupdateforkeys 
   - With that we would not add an additional library and not duplicate the logic
   - However it would not apply to `selectFeatureToggles` which might be userland responsibility again 

What do you think?

Kerumen:
Yes I think the right approach is to add it in `injectFeatureToggles` with the `onlyUpdateForKeys` from `recompose`.
`selectFeatureToggles` only provides the keys and can be used everywhere so it's the user to decide what to do. However in `injectFeatureToggles` we don't have any control.

[...]

PR #76, opened 2017-11-29
fix(react): so that injectFeatureToggles only updates for keys

tdeekens:
In #75 it was outlines that due to the fact that we treat flags as immutable (creating referentially different ones) we also re-render components wastefully which compose with `injectFeatureToggles`. 

This pull request attempts to reduce the wasteful renders by using `onlyUpdateForKeys` by `recompose`. It restricts re-renders only by the `flagNames` requested..

Closes #76


tdeekens:
@Kerumen please have a quick check. I think it's what we want?

Kerumen:
Yes! That's seems good!

PR #81, opened 2017-12-04
Release pool of version `5.0.x`

tdeekens:
💣  Breaking changes

- `FeatureToggled` and `render` (func) is now only invoked when feature matches
- `children` on the other-hand is invoked with `isFeatureEnabled` in any case
   - This matches behaviour in `react-router`
- Anywhere you used `variate` you should now use `variation`
   - We used the wrong terms here sorry for the confusion

🚬  Deprecations

- `FeatureToggled` got renamed to `ToggleFeature`
   - `FeatureToggled` will be supported for some time but using it will give you a `warning`
  
🌟  Features

- `SwitchFeature` is new component rendering the first matching `ToggleFeature`
    - It hope to ease so called multi variate testing by ensuring that only one feature is rendered at a time

```jsx
<SwitchFeature>
   <ToggleFeature flag="foo" variation="bar" toggledComponent={MyFeature}/>
   <ToggleFeature flag="foo" variation="baz" render={() => <p>Turned on</p>} />
   <ToggleFeature flag="foo" variation="bar">
       { ({ isFeautureEnabled } ) => <p>Turned on?: { isFeatureEnabled }</p> }
   </ToggleFeature>
   <ToggleFeature flag="foo" variation="foo" toggledComponent={MyFeature} untoggledComponent={DisabledFeature}/>
</SwitchFeature>
```

🥇  Fixes

- `ConfigureFlopflip` is now more mindful about not triggering wasteful renders
   - All of its unrelated state is extracted away from React to speed up your app
- `ConfigureFlopflip` and the underlying `FlagSubscription` had a hairy infinite configuration loop
   - The needed mix of `componentDidMount` and `componentDidUpdate` in pair with `configure` and `reconfigure` lead to a repeated brute force configuration of adapters which yielded a stack overlow
   - While fixing the bug we improved our internal state handling from naive boolean flags to more descriptive enums
- Wrapping of `displayNames` is now removed for all components
  - This makes snapshotting more consistent between releases
  - It eases testing with enzyme (a `wrapper.find('ToggleFeature` is enough)

💘 Tooling

- Dependencies have been updated again making sure `flopflip` is compatible with React@16.2


tdeekens:
There is nothing to add from my side. I plan to release this soon unless anyone would like to request another change (breaking changes welcome).

tdeekens:
Everything is publish under the `@next`-tag. I will do some integration testing tomorrow an then repoint the release to the `@latest` channel whenever ready.

tdeekens:
SNAFU: at least found a bug on the `@next` channel.

PR #82, opened 2017-12-04
feat: add FeatureSwitched component for feature switching

### LINKED ISSUE #79: RFC: Improved feature variate toggling through `FeatureVariate`
The idea is to simply feature variante toggling through declarative compound components using `FeatureToggled` and `FeatureVariante`.

```jsx
<FeatureToggled flag="signupForm">
   <FeatureVariate variate="version-1">
       <h1>Here we go new</h1>
   </FeatureVariate>
   <FeatureVariate variate="version-2">
       <h1>Here we go old</h1>
   </FeatureVariate>
</FeatureToggled>
```

tdeekens:
This adds a new component to the mix: the `FeatureSwitched`. The initial draft was noted in #79.

I went against `FeatureVariate` as I did not want to mix concerns with `FeatureToggled` which accepts only one child. Which might be wrong but is actually good to avoid additional markup.

The `FeatureSwitch` is quite simple. Its only purpose is to switch over it's `children` rendering the first component which matches. As a result the component is not incredibly useful as it doesnt compute a match itself. It only takes over ensuring that one feature is rendered. So maybe that makes it useful enough. 

```jsx
<FeatureSwitched>
  <FeatureToggled flag="foo" variate="bar">
     <p>Bar variate enabled</p>
  </FeatureToggled>
  <FeatureToggled flag="foo" variate="baz">
     <p>Baz variate enabled</p>
  </FeatureToggled>
  <FeatureToggled flag="foo" variate="barr">
     <p>Barr variate enabled</p>
  </FeatureToggled>
</FeatureSwitched>
```

The only downside of this is the duplication of the `flag` prop which we could lift up to `FeatureSwitched`. However, that comes at the cost of it not being able to be required on `FeatureToggled` due to the nature `cloneElement` and prop-type valiation. 

Closes #79.


tdeekens:
The "problem" of not having `flag` as a required prop could be solved by an `invariant`.

tdeekens:
Curious to what @Kerumen, @dferber90 or @luhmann think of this.

packages/react/modules/components/feature-switched/feature-switched.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const isEmptyChildren = children => React.Children.count(children) === 0;
+
+export default class FeatureToggled extends React.PureComponent {
```
luhmann: Wrong name for class?

I find `<FeatureSwitched>` a really weird name, because it is some grammar form that is not present tense or something "Look dude it is a feature but switched", why not:

* `<FeatureSwitch>` - elicits the image of a physical thing
* `<SwitchFeature>` - active imperative voice, nice and clear

packages/react/modules/components/feature-switched/feature-switched.spec.js
```diff
@@ -0,0 +1,94 @@
...
+FeatureComponent.displayName = 'FeatureComponent';
+
+describe('rendering', () => {
+  describe('with a feature match', () => {
+    let wrapper;
+
+    beforeEach(() => {
+      wrapper = shallow(
+        <FeatureSwitched>
+          <FeatureToggled isFeatureEnabled variate="foo">
```
luhmann: Now that it is enabled, I really wonder how widely used this will be without tracking on which Variant is more successful. 

Hmm but then again you can easily wire up your own tracking. Maybe add some documentation about this, the intended scope of the component, and maybe even a simple example in GA on how to implement tracking without bugging the maintainer for this lib about it? 😃

packages/react/modules/components/feature-switched/feature-switched.js
tdeekens: I went for past tense as `FeatureToggled` is. `SwitchFeature` seems also nice.

packages/react/modules/components/feature-switched/feature-switched.spec.js
tdeekens: Ha, yeah true that. Most adapters have the notion of tracking. E.g. LaunchDarkly with a `client.track("Signed up")`. You define goals there for A/B testing but I haven't used it. https://docs.launchdarkly.com/docs/running-ab-tests

packages/react/modules/components/feature-switched/feature-switched.js
luhmann: I kind of find `FeatureToggled` a weird name too 😃, but not weird enough for breaking changes and totally a personal preference. So don't mind me.

Kerumen: I agree with @luhmann, `SwitchFeature` is way better. And for `FeatureToggled`, `ToggleFeature` would have been nice. But anyway.. :p

dferber90: And while we're on it, `Variant` instead of `Variate` 🌺

dferber90:
> The only downside of this is the duplication of the `flag` prop which we could lift up to FeatureSwitched. However, that comes at the cost of it not being able to be required on FeatureToggled due to the nature cloneElement and prop-type valiation.

I'm for this and was about to recommend it until I saw the paragraph. Future versions of React will not have that problem anymore, since this will be supported. It's probably better to stay compatible with previous versions for now until some time has passed. Then we can upgrade with a breaking release.

packages/react/modules/components/feature-switched/feature-switched.js
tdeekens: You can't believe how happy I am for you all to engage on this. 

1. `FeatureToggled` is a strange name @luhmann @Kerumen 
   - Alright, lets just re-export a `ToggleFeature` and deprecate the other version
   - We can keep `FeatureToggled` for one or more major versions
2. Variant over Variate @dferber90 
   - I am just trying to follow LaunchDarkly's naming here as it was the first adapter and I feel it's the term used in the industry
   - http://blog.launchdarkly.com/tag/multivariate-feature-flags/

tdeekens:
> It's probably better to stay compatible with previous versions for now until some time has passed. Then we can upgrade with a breaking release.

Yeah that's kind of how I would take it too. Moreover, not lifting the `flag` prop up gives us the ability to `SwitchFeature` on multiple flags.

Kerumen:
But it seems redundant when you write the code. Maybe support both versions?

packages/react/modules/components/feature-switched/feature-switched.spec.js
Kerumen: And what about `variation`?

packages/react/modules/components/feature-switched/feature-switched.js
Kerumen: And what about `variation`?

tdeekens:
> Maybe support both versions?

Then the `prop` on `FeatureToggled` (soon to be `ToggleFeature`) can't be required anymore due to how prop-type validation and `cloneElement`.

packages/react/modules/components/feature-switched/feature-switched.js
tdeekens: Yeah, we can rename it if we want. I'll do that then. `variate` truly has caused confusion.

luhmann:
🎉

PR #84, opened 2017-12-05
BREADKING: renaming variate to variation

tdeekens:
The naming of `variate` has caused quite some confusion as a result @Kerumen suggested to rename it to `variation`. I hope this helps. Let me know what you think.


tdeekens:
Reading up on the matter the term was actually used falsely in our library. 

1. Multi variate feature toggling is just a grouping term for having a toggle which has variations
2. The variations is what you toggle on

This is good to go as a result.

PR #87, opened 2017-12-10
Add integration tests (mounting) for components

tdeekens:
> This pull requests plans to add integration tests for most of our API.

I fiddled with adding real puppeteer based integration tests but that seems a bit too much.


tdeekens:
When done, this might get coverage up to about 100%.

PR #93, opened 2017-12-27
Add `flow` for type checking

tdeekens:
> This pull request adds flow for type checking aiming at removing prop-types.

### Why?

First things first. Why flow, why type checking and why not TypeScript?

Type checking makes the library a bit more stable in my opinion. We won't use it as extensively as other libraries might. E.g. form libraries are a good example where type checking makes even more sense. However, generally I think libraries benefit from type checking.

TypeScript vs. Flow would be an endless debate. I think given the scope of type checks here it doesn't matter which one you pick. Even though I love TypeScript in general I have some things I dislike

1. Being a complete tool chain from language, to testing to linting
   - This might be beneficial for setup but I like Babel and ESLint doing their thing
2. Gradually adoptable and still only JavaScript
   - I am still writing JavaScript. I don't have to wait for things to be added to TypeScript
3. Integration into the React ecosystem
   - I feel Flow is better integrated into the React ecosystem
   - I would likely decide for TypeScript if I was writing an e.g. stream utility library
4. Flow is easier to integrate
   - TypeScript would include changing half the toolchain from removing Babel and ESLint to adding TSLint etc


packages/launchdarkly-adapter/modules/adapter/adapter.js
```diff
@@ -1,25 +1,56 @@
...
+type Client = {
+  identify: (nextUser: User) => void,
+  on: (state: string, () => void) => void,
+  on: (state: string, (flagName: FlagName) => void) => void,
+  allFlags: () => Flags | null,
+};
+type Flag = [FlagName, FlagValue];
+type Flags = { [FlagName]: FlagValue };
+type OnFlagsStateChangeCallback = Flags => void;
+type OnStatusStateChangeCallback = ({ [string]: boolean }) => void;
```
emmenko: Have you considered putting the types into its own file?

packages/launchdarkly-adapter/modules/adapter/adapter.js
```diff
@@ -1,25 +1,56 @@
...
+
+const adapterState: {
+  isReady: boolean,
+  isConfigured: boolean,
+  user: ?User,
+  client: ?Client,
+} = {
```
emmenko: What about an `AdapterState` type?

packages/launchdarkly-adapter/modules/adapter/adapter.js
```diff
@@ -34,23 +65,30 @@ const subscribeToFlagsChanges = ({ rawFlags, client, onFlagsStateChange }) => {
...
+const changeUserContext = (nextUser: User): void =>
+  adapterState.client && adapterState.client.identify
+    ? adapterState.client.identify(nextUser)
+    : undefined;
 
 // NOTE: Exported for testing only
-export const camelCaseFlags = rawFlags =>
-  Object.entries(rawFlags).reduce((camelCasedFlags, [flagName, flagValue]) => {
-    const [normalzedFlagName, normalzedFlagValue] = normalizeFlag(
+function entries<T>(obj: { [string]: T }): Array<[string, T]> {
```
emmenko: The exported comment is for the `camelCaseFlags` function right?

emmenko: Curious why have you "extracted" this function

packages/launchdarkly-adapter/modules/adapter/adapter.js
tdeekens: Yeap. I wanted to take a step at a time and see what the overlap actually is between packages. I don't want to have a `package/types` as that feels kind of strange when running in independent mode in lerna. WDYT?

packages/launchdarkly-adapter/modules/adapter/adapter.js
tdeekens: Yup, had that and struggled a bit with it. Especially as it's not reused. To me, right now, types make sense when you use them in generics for e.g. React.Component.

packages/launchdarkly-adapter/modules/adapter/adapter.js
tdeekens: My friend. You don't want to know. This cost me about 90 minutes and got me mad on a 🍷. It's due to the fact that flow seems to loose type information through the `Object.entries`. Reference on their repo [here](https://github.com/facebook/flow/issues/2174).

packages/launchdarkly-adapter/modules/adapter/adapter.js
emmenko: Hmm, depends what you want to do with the types, but no I would just simply have e.g. a `types.js` file in each package.

emmenko: But of course it would make sense if you are going to reuse those types. It's also perfectly fine to have types within the module itself if they are scoped there.

packages/launchdarkly-adapter/modules/adapter/adapter.js
emmenko: I know one thing or two about those "hours" 😅

packages/launchdarkly-adapter/modules/adapter/adapter.js
emmenko: Sure, I think often it comes down to how much you re-use that type.

packages/launchdarkly-adapter/modules/adapter/adapter.js
tdeekens: Yeah. I planned on starting - during the refactoring - to have them in each module. Then to move them to a package level `types.js`. The stop, no mono-repo wide types. As that feels like shared libraries in microservice land.

packages/launchdarkly-adapter/modules/adapter/adapter.js
tdeekens: Maddening. That's the only word coming to mind. However, there is little I honour more than all the hours people spend in building these great libraries. 👍

packages/launchdarkly-adapter/modules/adapter/adapter.js
tdeekens: a58d575

packages/launchdarkly-adapter/modules/adapter/adapter.js
tdeekens: a58d575

tdeekens:
CI should be humming along just fine now. Thanks for the suggesting @emmenko. Now comes some typing after fiddling with proper setup and CI intergration.

tdeekens:
Not sure if I will tackle adapters first and then all the react-* packages. Just to keep merges a bit smaller.

tdeekens:
Typing things like actions or even store enhancers is strange fun and fosters a really deeper understanding. You can not type what you don't understand.

tdeekens:
### Todo

- Scan for needlessly imported types (ESLint does not catch then)
- Fix render return values which can be more precise
- Fix FeatureTogglee fallback

packages/react-redux/modules/components/configure/configure.js
```diff
@@ -1,26 +1,37 @@
...
+} from '../../types.js';
+
 import React from 'react';
 import { connect } from 'react-redux';
 import { FlagsSubscription } from '@flopflip/react';
 import { updateStatus, updateFlags } from './../../ducks';
 
-export class Configure extends React.PureComponent {
+type Props = {
+  children: number,
```
tdeekens: Fix me

packages/react-redux/modules/components/configure/configure.js
```diff
@@ -1,26 +1,37 @@
...
+  shouldDeferAdapterConfiguration?: boolean,
+  defaultFlags?: Flags,
+  adapterArgs: AdapterArgs,
+  adapter: mixed,
+};
+type ConnectedProps = {
```
tdeekens: fn to void

PR #95, opened 2017-12-31
Add TypeScript definitions for `react-redux` and `react-broadcast`

tdeekens:
After having migrated all code to flow-type I wanted to offer TypeScript definitions files.

### Why TypeScript definitions?

Generally IDE support for TypeScript is quite advanced, many developers use VSCode and we can give inline type hints.

### Why not PR to definitely-typed?

I don't mind hosting the definitions linked by a `package.json` entry. Types don't get out of sync and don't have to be versioned in another repository. While developers would have to install `@types/flipflip/react-broadcast` mostly manually. All too much maintenance overhead for me instead of just keeping them here.

### Why not flow-types?

Flow has also has a `flow-typed` repository with types. However, I've seen a little amount of developers having flow setup in their IDE either installing types or expecting IDE integrations to give hints. I could probably also just link or re-export the internally held definitions or that already happens automatically with some setups.

### Why am I really doing this?

Learn about type systems.


tdeekens:
As a side note: the potential long term goal is to move over to TypeScript again when babel@7 supports removing the type annotations.

PR #96, opened 2018-01-02
Define bucklescript bindings (for react-broadcast)

emmenko:
This PR adds a new package `bs-flopflip` which is meant to contain all different [BuckleScript](https://bucklescript.github.io/) bindings to use those components in [ReasonML](https://reasonml.github.io/).

A binding is basically just the interop configuration between ReasonML <--> JS.

In this PR I focused on the bindings for the package `react-broadcast`. Please double check that the type definitions are exhaustive and the examples are correct.

> The Reason syntax should be fine, I haven't test the compiled JS though in a "real" react project.

If you want to see the output generated by BS, simply `cd` into the package and run `yarn start`. You will then see `*.bs.js` files next to the `*.re` files.

Feel free to comment in case you have any questions.

### DIFF
--- a/packages/bs-flopflip/README.md
+++ b/packages/bs-flopflip/README.md
@@ -7,16 +7,16 @@
 #### yarn
 
 ```bash
-$ yarn add bs-flopflip
+$ yarn add @flopflip/bs-flopflip
 ```
 
 #### bsconfig
 
-Add `bs-flopflip` to your `bs-dependencies`: **bsconfig.json**
+Add `@flopflip/bs-flopflip` to your `bs-dependencies`: **bsconfig.json**
 
 ```json
 "bs-dependencies": [
-  "bs-flopflip",
+  "@flopflip/bs-flopflip",
   ...
 ]
 ```
--- a/packages/bs-flopflip/README.md
+++ b/packages/bs-flopflip/README.md
@@ -1,0 +1,26 @@
+# `bs-flopflip`
+
+[BuckleScript](https://github.com/bucklescript/bucklescript) bindings for [flopflip](https://github.com/tdeekens/flopflip) packages - **WIP**
+
+## Install and setup
+
+#### yarn
+
+```bash
+$ yarn add bs-flopflip
+```
+
+#### bsconfig
+
+Add `bs-flopflip` to your `bs-dependencies`: **bsconfig.json**
+
+```json
+"bs-dependencies": [
+  "bs-flopflip",
+  ...
+]
+```
+
+## Usage
+
+_Coming soon_
--- a/packages/bs-flopflip/examples/react_broadcast.re
+++ b/packages/bs-flopflip/examples/react_broadcast.re
@@ -1,0 +1,48 @@
+/* Make all the module  */
+open ReactBroadcast;
+
+module UntoggledComponent = {
+  let component = ReasonReact.statelessComponent("UntoggledComponent");
+  let make = _children => {
+    ...component,
+    render: _self =>
+      <div>
+        (ReasonReact.stringToElement("At least there is a fallback!"))
+      </div>
+  };
+};
+
+/* ConfigureFlopFlip */
+module ExampleConfigureFlopFlip = {
+  let adapter: adapter = {
+    "configure": _args => Js.Promise.resolve(toAny("ok")),
+    "reconfigure": _args => Js.Promise.resolve(toAny("ok"))
+  };
+  let component = ReasonReact.statelessComponent("ExampleConfigureFlopFlip");
+  let make = _children => {
+    ...component,
+    render: _self =>
+      <ConfigureFlopFlip
+        adapter
+        adapterArgs={
+          "clientSideId": "123",
+          "user": Js.Nullable.return({"key": Js.Nullable.return("123")})
+        }>
+        <div />
+      </ConfigureFlopFlip>
+  };
+};
+
+/* ToggleFeature: https://github.com/tdeekens/flopflip#togglefeature */
+module ExampleToggleFeature = {
+  /* https://reasonml.github.io/reason-react/docs/en/component-as-prop.html */
+  let untoggledComponent = () => <UntoggledComponent />;
+  let component = ReasonReact.statelessComponent("ExampleToggleFeature");
+  let make = _children => {
+    ...component,
+    render: _self =>
+      <ToggleFeature flag="profile" untoggledComponent>
+        (ReasonReact.stringToElement("I might be gone or there!"))
+      </ToggleFeature>
+  };
+};
--- a/packages/bs-flopflip/package.json
+++ b/packages/bs-flopflip/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "name": "bs-flopflip",
+  "name": "@flopflip/bs-flopflip",
   "version": "0.1.0",
   "description": "BuckleScript bindings for flopflip",
   "license": "MIT",
--- a/packages/bs-flopflip/package.json
+++ b/packages/bs-flopflip/package.json
@@ -17,7 +17,7 @@
   },
   "homepage": "https://github.com/tdeekens/flopflip#readme",
   "scripts": {
-    "prebuild": "npm run clean",
+    "prebuild": "bsb -clean-world",
     "build": "bsb -make-world",
     "start": "bsb -make-world -w",
     "clean": "bsb -clean-world",
--- a/packages/bs-flopflip/package.json
+++ b/packages/bs-flopflip/package.json
@@ -1,0 +1,43 @@
+{
+  "private": true,
+  "name": "bs-flopflip",
+  "version": "0.1.0",
+  "description": "BuckleScript bindings for flopflip",
+  "license": "MIT",
+  "author": "Nicola Molinari <me@emmenko.org> (http://emmenko.org)",
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tdeekens/flopflip.git"
+  },
+  "bugs": {
+    "url": "https://github.com/tdeekens/flopflip/issues"
+  },
+  "homepage": "https://github.com/tdeekens/flopflip#readme",
+  "scripts": {
+    "prebuild": "npm run clean",
+    "build": "bsb -make-world",
+    "start": "bsb -make-world -w",
+    "clean": "bsb -clean-world",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "peerDependencies": {
+    "bs-platform": "^2.1.0"
+  },
+  "devDependencies": {
+    "@flopflip/react-broadcast": "*",
+    "bs-platform": "^2.1.0",
+    "reason-react": "^0.3.0"
+  },
+  "keywords": [
+    "feature-flags",
+    "feature-toggles",
+    "LaunchDarkly",
+    "client",
+    "reason",
+    "bucklescript",
+    "bindings"
+  ]
+}
--- a/packages/bs-flopflip/src/reactBroadcast.re
+++ b/packages/bs-flopflip/src/reactBroadcast.re
@@ -1,0 +1,131 @@
+type any;
+
+/* Since in ReasonReact we can't pass a component module as a prop (*),
+   we need to pass it as a function, which is still a React component.
+   https://reasonml.github.io/reason-react/docs/en/component-as-prop.html
+
+   (*) well, we kind of can 😅
+   [from the Discord channel:
+     "Sure you can, just give it a signature"
+     https://astrada.github.io/reason-react-playground/?reason=LYewJgrgNgpgBAFwJ4Ad4CUQQfAvHAbwCg45YE5gBDAaxgC44AKAPxSoQAt6BnBAJwCWAOwDmAGjhV+-KkiboYVHiGGKqAYwQA6fkq0BRWMBjCEASnNxcAPjjqVa-To0hgKVaYQKlj9Vu0+DhhYHh5JB1V-HWEQRQQqERgwAAV+EBRw+18o521NBEFVUJ5zAG4iAF8KolBIWHssHH5rQhIyGApXd08zVsinAsCEnBKAYTcPYS8mACJMbBh+WfL28kpaPGY9YTAlxiY66AwmmCtbbOVcob0CoxgTM3O7YlJSbQ-uqa9xdtI-uA7Pb8RgAfR4IQAZtYXgDSOt0ot+ABlEZbWbAJCo4KzCpvN5wyjgY6NRatV74t7rL69Cj4AbRYbBcaTWlzBY4FZ4ynwzobOitVjsLiSUEaTiCKBgHbPNo8-EfbQ06ZmX7ygnq0hA-ZwcFQmGE-EASAAPGBBAA3GyGt5Go0+K6DAJ8IRiAAqIHuj28iOa2JwcAA1IG4MLOJYbaRTQB6c1Wm2VQnVQmE7X8Q7Ehoc+CMbOrfGJnntZPJojRAAiAHkALKKXSmYEer1eADqgi4AEkwEx2ibs-xrQqPkwR0cs6dLAaeaa44PKUbSH3TqGOJxcLNo1RZnBo3PjYvsyuuOvowAjbe7m0m2OWmzmXvR-s2SSzFB6C2CGAAdy5QA
+   ]
+   */
+type reactClassForComponentProp = unit => ReasonReact.reactElement;
+
+type flagName = string;
+
+type flagVariation;
+
+type flag = {
+  .
+  "flag": flagName,
+  "variation": option(flagVariation)
+};
+
+type flags = Js.Dict.t(flagVariation);
+
+type adapterArgs = {
+  .
+  "clientSideId": string,
+  "user": Js.Nullable.t({. "key": Js.Nullable.t(string)})
+};
+
+type adapter = {
+  .
+  "configure": adapterArgs => Js.Promise.t(any),
+  "reconfigure": adapterArgs => Js.Promise.t(any)
+};
+
+/* Helpers */
+let optionToBool = optional =>
+  switch optional {
+  | Some(_) => true
+  | _ => false
+  };
+
+external toAny : 'a => any = "%identity";
+
+/* Module definitions (the actual bindings) */
+[@bs.module "@flopflip/react-broadcast"]
+external branchOnFeatureToggle :
+  (
+    ~flag: flag,
+    ~untoggledComponent: option(ReasonReact.reactClass)=?,
+    unit,
+    ReasonReact.reactClass
+  ) =>
+  ReasonReact.reactClass =
+  "branchOnFeatureToggle";
+
+[@bs.module "@flopflip/react-broadcast"]
+external injectFeatureToggle :
+  (
+    ~flagName: flagName,
+    ~propKey: option(string)=?,
+    unit,
+    ReasonReact.reactClass
+  ) =>
+  ReasonReact.reactClass =
+  "injectFeatureToggle";
+
+[@bs.module "@flopflip/react-broadcast"]
+external injectFeatureToggles :
+  (
+    ~flagNames: array(flagName),
+    ~propKey: option(string)=?,
+    unit,
+    ReasonReact.reactClass
+  ) =>
+  ReasonReact.reactClass =
+  "injectFeatureToggles";
+
+module ConfigureFlopFlip = {
+  [@bs.module "@flopflip/react-broadcast"]
+  external reactClass : ReasonReact.reactClass = "ConfigureFlopFlip";
+  let make =
+      (
+        ~shouldDeferAdapterConfiguration: option(bool)=?,
+        ~defaultFlags: option(flags)=?,
+        ~adapterArgs: adapterArgs,
+        ~adapter: adapter,
+        children
+      ) =>
+    ReasonReact.wrapJsForReason(
+      ~reactClass,
+      ~props={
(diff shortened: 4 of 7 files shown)


packages/bs-flopflip/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "private": true,
```
emmenko: It's marked as `private` on purpose, until you want to publish it. Up to you 😉

packages/bs-flopflip/README.md
```diff
@@ -0,0 +1,26 @@
+# `bs-flopflip`
+
+[BuckleScript](https://github.com/bucklescript/bucklescript) bindings for [flopflip](https://github.com/tdeekens/flopflip) packages - **WIP**
+
+## Install and setup
+
+#### yarn
+
+```bash
+$ yarn add bs-flopflip
```
tdeekens: I'd likely prefer having this renamed to `bs-flopflip-react-broadcast` (or whatever scoped package naming convention exists in Bucklescript land. So that we can also add a `bs-flopflip-react-redux` later on. WDYT?

packages/bs-flopflip/examples/react_broadcast.re
```diff
@@ -0,0 +1,48 @@
...
+
+/* ToggleFeature: https://github.com/tdeekens/flopflip#togglefeature */
+module ExampleToggleFeature = {
+  /* https://reasonml.github.io/reason-react/docs/en/component-as-prop.html */
+  let untoggledComponent = () => <UntoggledComponent />;
+  let component = ReasonReact.statelessComponent("ExampleToggleFeature");
+  let make = _children => {
+    ...component,
+    render: _self =>
+      <ToggleFeature flag="profile" untoggledComponent>
```
tdeekens: ...just to make sure: the `untoggledComponent` here expands to `untoggledComponent={untoggledComponent}` not to a boolean?

packages/bs-flopflip/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "name": "bs-flopflip",
+  "version": "0.1.0",
```
tdeekens: Just as a note: I'd prolly release this as 1.0.0 straight away. Not a big fan of 0.x.x things due to different matching with ranges.

packages/bs-flopflip/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "url": "git+https://github.com/tdeekens/flopflip.git"
+  },
+  "bugs": {
+    "url": "https://github.com/tdeekens/flopflip/issues"
+  },
+  "homepage": "https://github.com/tdeekens/flopflip#readme",
+  "scripts": {
+    "prebuild": "npm run clean",
+    "build": "bsb -make-world",
+    "start": "bsb -make-world -w",
```
tdeekens: `make-world`? Kind of a funny name.

packages/bs-flopflip/src/reactBroadcast.re
```diff
@@ -0,0 +1,131 @@
...
+   [from the Discord channel:
+     "Sure you can, just give it a signature"
+     https://astrada.github.io/reason-react-playground/?reason=LYewJgrgNgpgBAFwJ4Ad4CUQQfAvHAbwCg45YE5gBDAaxgC44AKAPxSoQAt6BnBAJwCWAOwDmAGjhV+-KkiboYVHiGGKqAYwQA6fkq0BRWMBjCEASnNxcAPjjqVa-To0hgKVaYQKlj9Vu0+DhhYHh5JB1V-HWEQRQQqERgwAAV+EBRw+18o521NBEFVUJ5zAG4iAF8KolBIWHssHH5rQhIyGApXd08zVsinAsCEnBKAYTcPYS8mACJMbBh+WfL28kpaPGY9YTAlxiY66AwmmCtbbOVcob0CoxgTM3O7YlJSbQ-uqa9xdtI-uA7Pb8RgAfR4IQAZtYXgDSOt0ot+ABlEZbWbAJCo4KzCpvN5wyjgY6NRatV74t7rL69Cj4AbRYbBcaTWlzBY4FZ4ynwzobOitVjsLiSUEaTiCKBgHbPNo8-EfbQ06ZmX7ygnq0hA-ZwcFQmGE-EASAAPGBBAA3GyGt5Go0+K6DAJ8IRiAAqIHuj28iOa2JwcAA1IG4MLOJYbaRTQB6c1Wm2VQnVQmE7X8Q7Ehoc+CMbOrfGJnntZPJojRAAiAHkALKKXSmYEer1eADqgi4AEkwEx2ibs-xrQqPkwR0cs6dLAaeaa44PKUbSH3TqGOJxcLNo1RZnBo3PjYvsyuuOvowAjbe7m0m2OWmzmXvR-s2SSzFB6C2CGAAdy5QA
+   ]
+   */
+type reactClassForComponentProp = unit => ReasonReact.reactElement;
+
+type flagName = string;
+
+type flagVariation;
```
tdeekens: This could be `string | boolean` depending on how union types work?

packages/bs-flopflip/src/reactBroadcast.re
```diff
@@ -0,0 +1,131 @@
...
+  | _ => false
+  };
+
+external toAny : 'a => any = "%identity";
+
+/* Module definitions (the actual bindings) */
+[@bs.module "@flopflip/react-broadcast"]
+external branchOnFeatureToggle :
+  (
+    ~flag: flag,
```
tdeekens: To my understanding this is slightly off and I had it wrong in my TypeScript definitions. Sorry about that. The actual implementation looks like [this](https://github.com/tdeekens/flopflip/blob/master/packages/react-broadcast/modules/components/branch-on-feature-toggle/branch-on-feature-toggle.js#L14).

packages/bs-flopflip/src/reactBroadcast.re
```diff
@@ -0,0 +1,131 @@
...
+    unit,
+    ReasonReact.reactClass
+  ) =>
+  ReasonReact.reactClass =
+  "branchOnFeatureToggle";
+
+[@bs.module "@flopflip/react-broadcast"]
+external injectFeatureToggle :
+  (
+    ~flagName: flagName,
```
tdeekens: This on the other hand is correct again. We don't need an object for config here as the evaluation of the value is up to the component being wrapped.

tdeekens:
A bit strange that the build fails due to running `yarn build` in the package when cleaning.

```txt
bs-flopflip: npm ERR! bs-flopflip@0.1.0 clean: `bsb -clean-world`
bs-flopflip: npm ERR! Exit status 2
bs-flopflip: npm ERR! 
bs-flopflip: npm ERR! Failed at the bs-flopflip@0.1.0 clean script.
bs-flopflip: npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
bs-flopflip: npm ERR! A complete log of this run can be found in:
bs-flopflip: npm ERR!     /home/circleci/.npm/_logs/2018-01-02T09_41_11_591Z-debug.log
bs-flopflip: npm ERR! code ELIFECYCLE
bs-flopflip: npm ERR! errno 2
bs-flopflip: npm ERR! bs-flopflip@0.1.0 prebuild: `npm run clean`
bs-flopflip: npm ERR! Exit status 2
bs-flopflip: npm ERR! 
bs-flopflip: npm ERR! Failed at the bs-flopflip@0.1.0 prebuild script.
bs-flopflip: npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
bs-flopflip: npm ERR! A complete log of this run can be found in:
bs-flopflip: npm ERR!     /home/circleci/.npm/_logs/2018-01-02T09_41_11_602Z-debug.log
lerna ERR! execute callback with error
lerna ERR! Error: Command failed: npm run build
lerna ERR! Error: spawnSync /home/circleci/flopflip/node_modules/bs-platform/lib/bsb.exe ENOENT
lerna ERR! npm ERR! code ELIFECYCLE
lerna ERR! npm ERR! errno 2
lerna ERR! npm ERR! bs-flopflip@0.1.0 clean: `bsb -clean-world`
lerna ERR! npm ERR! Exit status 2
```

packages/bs-flopflip/README.md
emmenko: It's not really necessary. I would prefer to have one package that contains all the bindings. That's why I named the file `reactBroadcast.re`. There will be one matching each package.

But of course if you prefer to have one for each package we can also do that. We can't however share e.g. the common type definitions.

packages/bs-flopflip/examples/react_broadcast.re
emmenko: Yes, it's called [punning](https://reasonml.github.io/guide/language/jsx/#punning).

emmenko: You can write it fully, then the "prettier" formatter ([refmt](https://github.com/facebook/reason/tree/master#installation-for-programmatic-usage)) will "pun" it 😉

packages/bs-flopflip/package.json
emmenko: Fine for me. We can also mark it as "alpha" or something if you want. I don't mind 😇

packages/bs-flopflip/src/reactBroadcast.re
emmenko: As far as I know it's discouraged to do that: https://reasonml.github.io/guide/language/variant/#variants-must-have-constructors

I haven't looked deeper to see what we can do in such cases.

Btw: if you just provide a type without a value, it's called an "abstract" type

>  It's a convenient way of specifying the relationship between external calls without knowing what the shape of the data is under the hood.

https://reasonml.github.io/guide/javascript/converting#clean-up-types-pass-2
https://reasonml.github.io/guide/language/module#signatures

emmenko:
> A bit strange that the build fails due to running yarn build in the package when cleaning.

Yeah, I have to double check it.

> Super nice learning exercise! I left 1-2 notes along the review. After we get this in I could try to "type out" react-redux which will be much a like.

Sure, we can also have a "session" together if you'd like.

packages/bs-flopflip/examples/react_broadcast.re
```diff
@@ -0,0 +1,48 @@
...
+        (ReasonReact.stringToElement("I might be gone or there!"))
+      </ToggleFeature>
+  };
+};
```
emmenko: Btw: the examples here are missing the HoCs. I still plan to add examples for those 😉

packages/bs-flopflip/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "private": true,
+  "name": "bs-flopflip",
```
emmenko: Also, do you want to scope this package?

packages/bs-flopflip/examples/react_broadcast.re
tdeekens: Nice. Read about refmt. Makes actually more sense then just assuming it to be true.

packages/bs-flopflip/src/reactBroadcast.re
tdeekens: Interesting and it makes sense from a type soundness perspective. To me it‘s a bit of a leaky business logic thing here. Flag variations, in almost all platforms with multi variate flags I know, are bools or strings. Which in turn leaks into code I guess. I guess I can not refactor this in accordance with the ˋdrawLineˋ and ˋdrawStringˋ example. I could just have two type contructors even if discouraged or work with abstract types (which feel like generics). So a ˋFlagVariation<T>ˋ. The latter might make sense as I actually never have care what concrete type it is. As long as I can === it. 

Thanks for the thought provocing links. I like this stuff.

packages/bs-flopflip/package.json
tdeekens: Yeah. If reason-land is fine with it Id prefer to scope it. For consistency and npm tidiness. I guess ´@flopflip/bs-react-broadcast´ would be nice.

packages/bs-flopflip/README.md
tdeekens: Ah. Disregard my comment below then. I fiddled with having shared TypeScript definitions but resorted to having much of them duplicated as configuration annoyed me a bit. If we can have one BS package, that would be nice. ´@flopflip/bs-flopflip‘ I guess. It would define and depend on both redux and broadcast variations I guess to define types for both. 

I will refactor shared flow- and TypeScript types to be a separate package also if needed.

emmenko: Btw: the "repository index" for reason packages is this one --> https://redex.github.io/

The cool thing is that you can already "submit" a package even if it's not published to npm yet, and it's marked as unpublished to indicate that is e.g. incomplete or experimental.
https://redex.github.io/packages/unpublished

packages/bs-flopflip/src/reactBroadcast.re
emmenko: Consider that this is just the interop interface so that the compiler can understand what's going on. In JS land is going to work as before (e.g. the `===` part).

It's also perfectly fine to start with really simple stuff and iterate on them later ([this page](https://reasonml.github.io/guide/javascript/converting) gives a good example on how you get more and more into Reason's type system).

emmenko:
> Yeah, I have to double check it.

Ok, I could reproduce the CI error locally when you "split" the steps like
1. `yarn install --pure-lockfile --ignore-scripts`
2. `lerna run --parallel build`

However, if you run `yarn`, everything works fine. I guess because some `postinstall` scripts are not being executed (e.g. https://github.com/BuckleScript/bucklescript/blob/master/package.json#L32).

What is the reason of installing the deps with `--ignore-scripts`?

tdeekens:
> `yarn install --pure-lockfile --ignore-scripts`

Is due to the fact that CircleCI otherwise died on the `lerna bootstrap` as lerna would run it in too many processes. The same problem we had on CI with Jest kind of. Maybe I'll just "hard-restrict" it all and refactor the build to run post install scripts. As that's likely the more important thing. Sorry for the inconvenience and thanks for finding out!

tdeekens:
BTW, rebasing this over `master` now might solve the CI issue caused by `--ignore-scripts`. I tried to solve it in #98.

emmenko:
@tdeekens seems to work! Thanks 🙏

tdeekens:
Glad it helped. So just silly me doesn't loose track listing what we still need here.

- [ ] Examples for HoCs as a "nice to have"
- [ ] Bindings for react-redux (from the same package)
   - Another PR or here?

Anything I missed?

emmenko:
> Bindings for react-redux

I think it's better to do a PR for each "package bindings".

emmenko:
Also, I have to change the HoC bindings, it doesn't really work like this in Reason. Working on it...

emmenko:
It's a bit more complicated than I thought. I think I'm going to remove the bindings for the HoCs from this PR and open a separate one.

You can read this blog post that explain quite well the problematic and how you can "make it better".
https://medium.com/@_gabrielrubens/binding-a-react-hoc-to-reasonml-children-as-function-d2688e4edaae

In general the approach is to have a "declarative" component with a FaaC. In the blog post there is an example of `withState` from `recompose`.
The bindings for Apollo also go with that approach (to implement the `graphql`HoC)https://github.com/apollographql/reason-apollo

In general, I actually like more the declarative approach, combined with pattern matching results in a really nice API.

What I can imagine for the `branchOnFeatureToggle` HoC would be something like this:

```
<BranchOnFeatureToggle flag="foo" variation="1">
  (
    branch => {
      switch branch {
        | Left => /* The component to be toggled */
        | Right => /* The fallback component */
      }
    }
  )
</BranchOnFeatureToggle>
```

tdeekens:
Interesting. Will have a read. Always open to improve things. Especially when it is around the area of from FaaC to HoC and Render Prop.

tdeekens:
I guess, from my current understanding, it's something which could also feedback into the library itself. Most components could become more declarative in a FaaC/Render-Prop fashion in e.g.

```js
<InjectFeatureToggle flag="foo">
  {({ featureToggle }) => ...}
</InjectFeatureToggle>

<InjectFeatureToggles flags=['foo', 'bar']>
  {({ featureToggles }) => ...}
</InjectFeatureToggle>

<BranchOnFeatureToggle flag="foo" variation="baz">
  {({ match }) => ...}
</InjectFeatureToggle>
```

while all HoCs would just render those as in:

```js
export default injectFeatureToggle('bar', 'propKey') => WrappedComponent => 
    <InjectFeatureToggle>{({ featureToggle }) => React.cloneElement(WrappedComponent, { [propKey]: featureToggle }}<InjectFeatureToggle>
```

Kind of I guess.

emmenko:
Yeah! I think it would be good in general to have a “declarative” version of the HoC 👍

tdeekens:
I think most of that can be achieved with `ToggleFeature` anyway. Currently the library defined and uses the primitives the other way round. The HoCs are used to build up to `ToggleFeature`.

emmenko:
I've been thinking about this a bit more, also about what you said, and I believe that we only need to bind the `ToggleFeature` component.
Even in JS land, if we were to avoid using/having HoCs, we would probably end up using only `ToggleFeature` anyway (unless I'm forgetting about other use cases).

In any case, since in Reason we "can only" use components, it would make sense to have one.

What do you think?

Furthermore, in JS there are different ways of rendering this `ToggleFeature` component: with children, with render, with props, with FaaC 🙃
If we were to build this component in Reason, and given the fact that [passing a component module as a prop does not work straight away](https://reasonml.github.io/reason-react/docs/en/component-as-prop.html), I would probably stick to only one way of rendering it.

```ocaml
<ToggleFeature flag="foo" variate="bar">
  (
    (match: bool) => {
      switch match {
        | true => /* The component to be toggled */
        | false => /* The fallback component, or null */
      }
    }
  )
</ToggleFeature>
```

---
Anyway, I removed the HoC bindings from this PR for now.

tdeekens:
> I believe that we only need to bind the ToggleFeature component. [...] What do you think?

Yeah, was walking around town and actually thinking that. It's what stuck in my head with "the primitives are upside down". Essentially you can build HoCs from a declarative component. Which feels more right to me anyways. "Sadly" the project builds declarative components using HoCs which might be the "wrong" way around.

For the Reason bindings I think it makes total sense to just use `ToggleFeature` and maybe not even bind any HoCs. I guess it's what you were after after all. I don't think all bindings necessarily need the same API. A subset should be fine. As long as they have the same - super far future speaking - bindings across one UI platform.

emmenko:
> Essentially you can build HoCs from a declarative component. Which feels more right to me anyways. "Sadly" the project builds declarative components using HoCs which might be the "wrong" way around.

I see. Do you intend to "refactor" this internally then or what?

> I don't think all bindings necessarily need the same API.

That's correct.

---
I'll see if I can change the exposed reason API of `ToggleFeature` like the example above. The way it is right now it matches 1:1 the props of the JS API, so it should work but it doesn't have the "ideal" API.

tdeekens:
> I see. Do you intend to "refactor" this internally then or what?

"or what?" 😄. I would like to and will see when I've got time. I think I would have to

1. Refactor `ToggleFeature` to accept `flags: Flag | Array<Flag>`
   - Then I could connect/subscribe `ToggleFeature` itself to flag(s)
   - While using it for `injectFeatureToggle` and `injectFeatureToggles`
   - Maybe there would be a `ToggleFeatures` to not have the component "do too much" (see below)
2. Refactor `branchOnFeatureToggle` to use `ToggleFeature`
   - Unsure how the `isFeatureEnabled` of `ToggleFeature` would behave is it differs to `Flag` and `Array<Flag>`
3. Refactor `injectFeatureToggle` to compose using `ToggleFeature`

To me the value of this is mostly code architecture wise (apart from React DevTools render trees looking alike across components). It "breaks" a bit with the package split of the library which is a 1-to-1 mapping of `react` to `react-{redux, broadcast}`. I also fear the changes to the tests a bit which might take some time.

Hm, I might not see the forest for the trees but the change being mostly cosmetic I'd maybe focus on other things while keeping it on the "train-side/bus-ride" task list 🐱.

tdeekens:
Looks good to me now. Sorry for the little derail about the direction and structure of things in this PR! 

This would be ready to go, right? Just a last little question regarding naming. Do you think `@flopflip/bs-flopflip` is a good name (duplicated library name) or should I/we go with `@flopflip/react-bs` or `@flopflip/react-bucklescript` or even `@flopflip/react-reason`? Unsure about terminology here. I would obviously carry out the renaming...

emmenko:
Actually hold on a sec. Let me gather some thoughts and get back to you later.

emmenko:
Ok I've got some time now to write this down.

So my initial idea/approach was to provide bindings for each individual package + exports, that's usually how you would do those things.

However after the discussion we had (see above) it's clear to me that we need to do things a bit differently. The question then is not much "what do we need to bind" but rather "what modules should we expose in Reason-land".
That means that some components might be bind 1:1 (e.g. `ConfigureFlopFlip`) but others need to be exposes in a way that the API makes sense in Reason (e.g. `ToggleFeature` vs _HoC_).

This leads us to the package/folder structure. Again, my initial idea was to provide a _file_ matching each package, however given the assumptions above I would propose something different: **each module/file represents a component that we want to expose**.
Therefore the folder structure should look something like:

```
- src
  - configureFlopFlip.re
  - toggleFeature.re
  - ...
  - flopFlipTypes.re
```

This would answer also your question about the naming. The package can simply be called `@flopflip/reason` or `@flopflip/reason-bs`.

What do you think?
I would probably start a new PR then.

emmenko:
I opened a new PR with the new approach.

Further discussions should still be done here.

tdeekens:
Thanks for your thoughts. All sounds very reasonable. Especially the module/file to component matching you propose. I'll check out the subsequent PR. Should be close this one then?

emmenko:
I guess, unless we have other things to talk about 😄

tdeekens:
Nope 😄. Unless we want to chat about the weather.

PR #97, opened 2018-01-03
Add pre-commit hooks (commitlint, lint-staged)

emmenko:
Not sure if you are interested in having those 😇

image

image


tdeekens:
Thanks a lot for this. Personally I never felt a super high urge for this so I never focussed on adding it. We should likely also add flow-type checking as a hook? Shouldn't we? What do you think?

emmenko:
> Personally I never felt a super high urge for this so I never focussed on adding it.

Yeah, not a big deal. I think it's nice to enforce some conventions especially for new contributors 🤓

> We should likely also add flow-type checking as a hook?

Sure thing, we can add as many things as we want. We could also run prettier if you'd like. 💅

tdeekens:
Bloep. GitHub's merge conflicts view is amazing but it's still saying that the conflicts exist. Even after resolving them.

emmenko:
That should be good now. @tdeekens do you need me to add/change anything specific?

tdeekens:
Nope. I'll merge this and add the flow checking myself. Thanks!

tdeekens:
Maybe I'll also add a note to the contributing section. If there are more contributors at some point wondering about the convention.

PR #100, opened 2018-01-03
Add flow-type checking and prettier formatting to `lint-staged`

tdeekens:


.nvmrc
```diff
@@ -0,0 +1 @@
+8
```
tdeekens: Can't harm to have an .nvmrc.

package.json
```diff
@@ -12,6 +12,8 @@
     "format": "npm run format:md && npm run format:js",
     "format:md": "prettier --parser markdown --write '*.md'",
     "format:js": "prettier --write '**/modules/**/*.js' '**/demo/**/*.js'",
+    "format:ts": "prettier --parser typescript --write '*.ts'",
+    "fix:eslint": "eslint --fix --format=node_modules/eslint-formatter-pretty",
```
tdeekens: Better extract this into a run-script while also adding TypeScript formatting through prettier.

package.json
```diff
@@ -103,7 +105,24 @@
     "packages/*"
   ],
   "lint-staged": {
-    "packages/**/demo/**/*.js": ["eslint --fix --format=node_modules/eslint-formatter-pretty", "git add"],
-    "packages/**/modules/**/*.js": ["eslint --fix --format=node_modules/eslint-formatter-pretty", "git add"]
+    "demo/**/*.js": [
```
tdeekens: Split's this up into more specific sections to format and lint only where needed. This means that we can: auto-format JS and TS while also formatting documentation and e.g. type-check using flow before commiting.

PR #102, opened 2018-01-04
Remove unnecessary `lerna bootstrap` command

emmenko:
When using yarn workspaces, it's not necessary to run `lerna bootstrap`
as it might get into an infinite loop when execute on postinstall step.
See https://github.com/lerna/lerna/issues/1125 and related issues

In case the prepublish and prepare script hooks need to be executed, you
can simply trigger those "manually" in the postinstall script.
```
"postinstall": "check-node-version --package --print && lerna run prepublish && lerna run prepare"
```

Side note: is the `build` step supposed to be executed separately from the "installation" step?

```
$ yarn
$ yarn build
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,10 +16,6 @@ aliases:
     # Ignoring scripts (e.g. post-install) to gain more control
     # in the jobs to only e.g. build when explicitely needed.
     command: yarn install --pure-lockfile
-
-  - &bootstrap
-    name: Bootstraping
-    command: yarn bootstrap
   
   - &lint
     name: Linting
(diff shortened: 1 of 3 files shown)


tdeekens:
Thanks! Must be a left over from the good-old yarn beta workspaces or something.

> Side note: is the build step supposed to be executed separately from the "installation" step?

Yes, I recently changed it with the CI improvements to BuckleScript. It started getting on my nerves a little. With CI building all the time anyway and building running as a pre run-script to `yarn release` I never felt it served a clear purpose for me.

emmenko:
That was quick! 😄 🚀

PR #106, opened 2018-01-05
Add package to hold Bucklescript bindings for ReasonML

emmenko:
> This PR supersedes #96 ([read the reasons behind it](https://github.com/tdeekens/flopflip/pull/96#issuecomment-355568791)).

The PR contains:
* new package setup (for bucklescript)
* the package name `@flopflip/reason-bindings` is still a **proposal**, it's not 100% decided yet
* a `example` folder that is meant to contain usage examples for each binding
* a module `FlopFlip__Types.re` that shares the types across the different components 
* bindings for `ConfigureFlopFlip` (react-broadcast)
  * a module `FlopFlip_ReactBroadcast__ConfigureFlopFlip.re` that directly exposes the react component (file naming is important!)
* (_bindings_) for `ToggleFeature` (react-broadcast)
  * a module `FlopFlip_ReactBroadcast__ToggleFeature.re` that implements a bit different API than its JS counterpart

> Read about [namespacing](https://github.com/tdeekens/flopflip/pull/106#issuecomment-355862017).
 
Other bindings will be provided in separate PRs

---
NOTE: we don't provide 1:1 bindings with each one of the components from each package. We only provide bindings **needed to work with the library** in Reason-land.


tdeekens:
Nice, I like that we just offer bindings for parts of flopflip which make sense in Reason. All looks good to me. Thanks! I guess picking up the plan of the previous PR we just need to port `ToggleFeature`. I guess it's worth checking out `bs-react-router` as it's essentially the same API patterns.

emmenko:
The RR bindings are actually very basics, as they are just binding components (I know because I’m also contributing to it 😉).

For our case, if we go with the approach I proposed, is a bit more complicated as it’s not a 1:1 binding. We can maybe discuss this in the next days in person, just to have a better understanding 🤓

Another question: are there any other components that we need to port? If so should we make a list of them?

tdeekens:
> are there any other components that we need to port?

That depends on the package. For `react-broadcast` we need

- [x] `ConfigureFlopflip`
- [ ] `ToggleFeature`
- [ ] `SwitchFeature` (the same as `Switch` in react-router)

that's all a user of the lib needs. Setup and config using `ConfigureFlopflip` and toggling using `ToggleFeature`. Some other intermediary components would be used behind the scenes such as `FlagSubscription` to integrate with `react-broadcast` but I guess we don't need to bind then?

for `react-redux` the minimal set would be the same with some additional parts (depending on if we need to bind them).

- [x] `ConfigureFlopflip`
- [ ] `ToggleFeature`
- [ ] `SwitchFeature`
- [ ] `flopflipReducer` (to integrate into the store)
- [ ] `FLOPFLIP_STATE_SLICE ` (constant key of the reducer)
- [ ] `flopflipReducer` (to integrate into the store)
- [ ] `createFlopFlipEnhancer` (for people who don't want to configure using `ConfigureFlopflip`)

I would assume we would have to bind most setup things for redux. Obviously not all in one PR if that makes things easier.

emmenko:
Understood!

For the redux part I would actually only bind the components.
The “reducer” parts I guess are used to configure redux. So unless you are also using redux in reason (which I haven’t seen a library yet) you will still keep that part in JS, hence there is no need to bind those things yet and I would suggest to get back to those once there is a need for it.

WDYT?

tdeekens:
Sure makes total sense. I did strike them from the above list

emmenko:
Short update (see last commit https://github.com/tdeekens/flopflip/pull/106/commits/7007a4b9a2b0ab689edba20a1f4945b528154769): we need to introduce namespacing, especially because within the library itself there are different exports of the same component (e.g. `ConfigureFlopFlip`) from different packages.

So at the end you would use the modules like:

```reason
<FlopFlip_ReactBroadcast.ConfigureFlopFlip />

<FlopFlip_ReactRedux.ConfigureFlopFlip />
```

tdeekens:
Sorry for not being on top of this. The namespacing manual namespacing makes total sense until the scoped package namespacing is improved.

packages/reason-bindings/src/FlopFlip_ReactBroadcast__ToggleFeature.re
```diff
@@ -0,0 +1,40 @@
...
+  let make = (~channel: string, children) =>
+    ReasonReact.wrapJsForReason(
+      ~reactClass,
+      ~props={"channel": channel},
+      children
+    );
+};
+
+[@bs.module "@flopflip/react"]
+external isFeaturedEnabled :
```
tdeekens: IsFeatureEnabled.

packages/reason-bindings/src/FlopFlip_ReactBroadcast__ToggleFeature.re
```diff
@@ -0,0 +1,40 @@
+open FlopFlip__Types;
+
+module Subscriber = {
+  [@bs.module "react-broadcast"]
+  external reactClass : ReasonReact.reactClass = "Subscriber";
```
tdeekens: Do we have to bind this component as it‘s used by ToggleFeature?

packages/reason-bindings/examples/example_configure_flopflip.re
```diff
@@ -1,4 +1,4 @@
-open FlopFlipTypes;
+open FlopFlip__Types;
```
tdeekens: Silly question: is the double underscore intentional?

packages/reason-bindings/src/FlopFlip_ReactBroadcast__ToggleFeature.re
emmenko: Well yes, the ToggleFeature implementation here is different from the original one.

packages/reason-bindings/src/FlopFlip_ReactBroadcast__ToggleFeature.re
emmenko: Ups

packages/reason-bindings/examples/example_configure_flopflip.re
emmenko: Yes, apparently is some kind of convention for namespacing

tdeekens:
I am closing this. I am very thankful for the contribution but I lost track of the progress and status nor has anybody else demanded this. I would love to re-open this once work on this is active again.

PR #235, opened 2018-03-07
Add reference when non normalised flag names are encountered

tdeekens:
> This pull request aims to to target a discussion around unexpected behaviour of the library when passing non-normalised flags.

I am still not too sure if I would like to add a `skipFlagNormalization` prop to `<ConfigureFlopflip> yet. In mean time - and even after - I think it's worthwhile to provide descriptive warnings to the user whenever the library encounters a situation where it might behave other than anticipated.

This means:

- Whenever a passed flag - either in `react-broadcast` or `react-redux` ends up in `isFeatureEnabled`
   - and the flag name does not seem normalized
   - then we warn the user with a reference to the respective readme section

Again, this same behaviour should exist whenever a `skipFlagNormalization` prop is added. However I see some problems with that still which can be discussed on the respective issue and elaborated on with PoCs.

### DIFF
--- a/packages/react/modules/helpers/is-feature-enabled/is-feature-enabled.js
+++ b/packages/react/modules/helpers/is-feature-enabled/is-feature-enabled.js
@@ -1,3 +1,5 @@
+import camelCase from 'lodash.camelcase';
+import warning from 'warning';
 import { DEFAULT_FLAG_PROP_KEY } from '../../constants';
 
 /**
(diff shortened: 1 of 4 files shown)


tdeekens:
> Love how simple it was to add this!

Yeah me too. The only thing I don't really like is the implicit assumption of `lodash.camelCase` to equal flag normalization. Maybe it can be extracted. On the other hand I don't value shared utils.

PR #244, opened 2018-03-09
fix(ts-bindings): add flag, variation, isFeatureEnabled to optional for toggle components

trisrael:
The [documentation for ToggleFeature](https://github.com/tdeekens/flopflip#togglefeature) states to support and required use of `flag` and `variation` properties but the TS defns are missing them.

Add these two properties for the `Toggle` components in react-broadcast & react-redux, leave original `Toggle` components from react package alone as it is correct.

### DIFF
--- a/packages/react-broadcast/typings/index.d.ts
+++ b/packages/react-broadcast/typings/index.d.ts
@@ -2,5 +2,10 @@
 // TypeScript Version: 2.x
 
 declare module '@flopflip/react-broadcast' {
+  import Flag from '@flopflip/types';
   export * from '@flopflip/types';
+
+  export interface ToggleComponentProps extends Flag {
+    isFeatureEnabled?: boolean;
+  }
 }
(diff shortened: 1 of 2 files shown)


trisrael:
@tdeekens need to some testing on this, if there are notes somewhere on building the packages so that I can use them with `npm link` lemme know

tdeekens:
Thanks. You can build things by running the build run script either for all or a scoped package via lerna. Linking things should not be affectee by that. Hope that helps.

packages/react-broadcast/typings/index.d.ts
```diff
@@ -2,5 +2,10 @@
 // TypeScript Version: 2.x
 
 declare module '@flopflip/react-broadcast' {
+  import Flag from '@flopflip/types';
   export * from '@flopflip/types';
+
+  export interface ToggleComponentProps extends Flag {
+    isFeatureEnabled?: boolean;
```
tdeekens: Sorry for the late response. I don't think this is correct. `ToggleComponent` receives a `flag` and a `variation` this `variation prop goes through `isFeatureEnabled` which "translates" it to a `boolean` which then hits `ToggleComponent` in the react package. This way we can share most of the component's logic.

tdeekens:
Added a PR aiming at fixing the typing. Will merge and release that soon which should close this one. Thanks for reporting and all.

PR #270, opened 2018-03-29
Reconfigure `launchdarkly-adapter` on non deep equality of users

### LINKED ISSUE #268: [@flopflip/launchdarkly-adapter] Specify different key to trigger `reconfigure`
Hey there - I'm not sure if I have a unique use case or not, but here goes:

**Issue:**
- We use a user's `email` as their `key` with LaunchDarkly
- We have some flags that we configure on a per user basis / user segments. However, we also have flags that are dependent on the custom context of what organizationId and eventId the user is currently viewing.
- This works great when forcing a new page load when the user jumps from event to event. However, when sticking with SPA routing, I can't figure out how to force the LaunchDarkly adapter to `reconfigure` because technically the user's key (`email`) is not changing.

**Solution I'm thinking of:**
- My thinking is to allow a different `key` to be provided to the adapter for it to check against instead of using `key`. For example, if I pass a prop of `reconfigureKey`, the adapter will use that for it's internal validation. If `reconfigureKey` is not provided, it falls back to `key`.

**Example solution:**
```
  if (adapterState.user && (adapterState.user.configureKey !== user.configureKey || adapterState.user.key !== user.key)) {
    adapterState.user = ensureUser(user);

    return changeUserContext(adapterState.user);
  }
```

Reference line: https://github.com/tdeekens/flopflip/blob/master/packages/launchdarkly-adapter/modules/adapter/adapter.js#L75

I wanted to make sure I'm not blind to any other provided solutions. If so, please enlighten me! 🤓  Otherwise, let me know if you'd be interested in this as a PR - else, I can fork.

Appreciate your hard work!

tdeekens:
> Previously the user `key` property was used to determine reconfiguring the adapter.

This might not be intended as other user properties can also change about which LD needs to be notified in case they change on the client.

This pull request changes the equality check some key equality to a deep equals check via lodash (note we use lodash for all utils).

Closes #268

### DIFF
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.js
@@ -9,6 +9,7 @@ import type {
   OnStatusStateChangeCallback,
 } from '@flopflip/types';
 import warning from 'warning';
+import isEqual from 'lodash.isequal';
 import { initialize } from 'ldclient-js';
 import camelCase from 'lodash.camelcase';
 
@@ -171,7 +172,7 @@ const configure = ({
 
 const reconfigure = ({
   clientSideId,
-  user,
+  user: nextUser,
   onFlagsStateChange,
   onStatusStateChange,
 }: {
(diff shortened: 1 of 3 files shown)


tdeekens:
Sometimes I wonder why it was not like this in the firsr place. Wondering if I missed something but I cant think of anything.

tdeekens:
Will publish before Monday.

PR #280, opened 2018-03-31
Add `ReconfigureAdapter` component

### LINKED ISSUE #277: [RFC]: Add `ReconfigureFlopflip` component
> There might be a case for a `ReconfigureFlopflip` component for this library.

### Introduction

Flopflip is generally configured by `ConfigureFloplip` by passing an `adapter`, `adapterArgs` among other things. Depending on the underlying adapter the user context can change. This is handled automatically by the `launchdarkly-adapter` as of late. Whenever any of the user props change the underlying adapter will reconfigure through `ConfigureFlopflip`. This may yield a new set of flags depending on your flag configuration.

### Case for the addition

A project or mine has a setup in which `ConfigureFlopflip` is rendered very high up the tree. Some of the user properties however as part of the `adapterArgs` are only known further down the component tree. This is instance can hold for any project with RR4 where a `<Route>` quite low in the tree. Imagine anything of that `match` prop should end up in the `user` of `adapterArgs`. You're stuck having to pass down an quite imperative setter function e.g. `setProjectKey` which somehow sideways changes the user context. If this is possible depends on the adapter. The `ldAdapter` exposes a `changeUserContext` which shallowly merges in any new passed props of a `nextUser`. This might seem doable at first but is not really declarative.

The same could be solved by a `ReconfigureFlopflip` which only receives e.g. two props `nextAdapterArgs` and a `exact` prop (defaulting to `false`). Depending on if `exact` is `true` the `nextAdapterArgs` are shallowly merged or overwritten.

### Example

```jsx
<Route
    exact={false}
    path="/:projectKey"
    render={routerProps => (
       <React.Fragment>
           <MyRouteComponent />
           <ReconfigureFlopflip exact={false} adapterArgs={{ projectKey: routerProps.projectKey }} />
       </React.Fragment>
    )}
/>
```

### Changes required

1. Adding the `ReconfigureFlopflip` component itself
2. The `ConfigureFlopflip` has to pass the adapter (or only `reconfigure`) onto the context of flopflip
3. Potentially use `calculateChangeBits` for the flags and `reconfigure` on the context to subscribe to only needed context state

tdeekens:
> This pull request aims to add a `ReconfigureAdapter` component to declaratively update the user context.

- Closes #277 

### Why TL;DR (please check issue)

You can do this anywhere in the tree

```jsx
<ReconfigureFlopflip exact={false} user={{ projectKey: routerProps.projectKey }} />
```

#### Use "legacy" context API

Could have used the "old" context API but it will likely be deprecated in 16.4. As a result the `react` package now also depends on `create-react-context` (as a polyfill) which will increase bundle size.

#### Merge `flags` and `adapter` contexts

This was attempted but felt inconvenient. In the `react-redux` implementation `flags` never live "on the context (well they do - not on ours)" and it felt confusing to have a context value for them. The implementation would have used `calculateChangedBits` and `observedBits` to have consumer target context state without wasteful re-renderings. Moreover, this is close to impossible to lift `flag` state up into the `react` package as it's management is in either `react-broadcast` or `react-redux` with the corresponding state management solution.

#### Use `create-react-context` in `react` (package)

This is how it's implemented. The `configure-adapter` module buts a `reconfigure`-fn onto the context. The `with-reconfigure`-HoC (do not hit me) injects it into the `reconfigure-adapter` which uses lifecycle hooks to call up to `configure-adapter`. `configure-adapter` in turn does not solely configures through `props.adapterArgs` anymore but buts them onto local component state. Whenever `reconfigure-adapter` calls to `reconfigure` `configure-adapter` will internally invoke `setAdapterArgs` which will update that component state. This in turn will trigger a `componentDidUpdate` which will invoke `reconfigure` on the underlying adapter whenever it's initialised. 

This way the `adapterArgs` are on the component state and things remain as "deterministic" as possible. As any underlying (down-tree) update of `adapterArgs` could get the adapter and the props of `configure-adapter` out of sync. 

#### Additional Improvments

- Use `lodash.memoize` to calculate `adapterArgs` to not re-render components
- Rebrand `with-flag-subscription` to `with-flags` and move it to `configure-adapter`
   - This way `react-broadcast` has a context HoC (for flags) in `configure-adapter` while `react` has the reconfiguration context HoC there


tdeekens:
CI is right about the failing snapshots. I can't update them locally however. Strange.

tdeekens:
Wohooo. This actually does work. Verified again in the `demo`. However, it still doesn't "really" solve the problem of `reconfiguring` through `ReconfigureFlopflip` _before_ the adapter is actually configured. Even if the `localstorage-adapter` configures on the next tick (`Promise.resolve`) the `ReconfigureFlopflip` will render and `didMount` before even. Not super sure how to solve this.

1. Add a `reconfigurationQueue` with properties which need merging (in `ReconfigureFlopflip`)
   - Work the queue every `props.queueWorkDelayMs` 
2. Just `catch` into the `reconfigure` call and `setTimeout` another round for reconfiguration

I guess the expected behaviour with e.g. 3x `ReconfigureFlopflip` in an app is that all additional `user` props are merged in.

WDYT @emmenko?

tdeekens:
Wohooo. This does work as verified in the `demo`. The careful ping pong from `reconfigureOrQueue` to wait for the `configure` to run. To then check the `pendingAdapterArgs` to then set state and `didUpdate which does not retrigger another set state.

packages/react/modules/components/configure-adapter/configure-adapter.js
```diff
@@ -27,19 +22,98 @@ type Props = {
...
+type AdapterReconfigurationOptions = {
+  exact?: boolean,
+};
+type AdapterReconfiguration = {
+  adapterArgs: AdapterArgs,
+  options: AdapterReconfigurationOptions,
+};
+type ReconfigureAdapter = (
+  adapterArgs: AdapterArgs,
+  { exact?: boolean }
```
tdeekens: Memo: change to reconfig options.

packages/react-broadcast/modules/components/inject-feature-toggles/inject-feature-toggles.js
```diff
@@ -21,6 +21,6 @@ export default <RequiredProps, ProvidedProps>(
 ) => (WrappedComponent: ComponentType<$Diff<RequiredProps, ProvidedProps>>) =>
   compose(
     setDisplayName(wrapDisplayName(WrappedComponent, 'injectFeatureToggles')),
-    withFlagSubscription(ALL_FLAGS_PROP_KEY),
+    withFlags(ALL_FLAGS_PROP_KEY),
```
tdeekens: Memo: default on HoC site not here.

packages/react/modules/components/configure-adapter/with-reconfigure.js
```diff
@@ -0,0 +1,33 @@
...
+import type { Adapter } from '@flopflip/types';
+
+import React, { PureComponent, type ComponentType, type Node } from 'react';
+import { wrapDisplayName } from 'recompose';
+import { AdapterContext } from './configure-adapter';
+
+type RequiredProps = {};
+type ProvidedProps = {};
+
+const withReconfigure = (propKey: string = 'reconfigure') => (
```
tdeekens: Memo: rename to withReconfiguration.

packages/react/modules/components/configure-adapter/with-reconfigure.js
```diff
@@ -0,0 +1,33 @@
...
+  Component: ComponentType<RequiredProps>
+) => {
+  class EnhancedComponent extends PureComponent<
+    $Diff<RequiredProps, ProvidedProps>
+  > {
+    static displayName = wrapDisplayName(Component, 'withReconfigure');
+    render(): Node {
+      return (
+        <AdapterContext.Consumer>
+          {reconfigure => (
```
tdeekens: Memo: memoize the API.

packages/react/modules/components/configure-adapter/with-reconfigure.spec.js
```diff
@@ -0,0 +1,35 @@
...
+  beforeEach(() => {
+    const Component = withReconfigure('foo-prop-key')(TestComponent);
+    wrapper = shallow(<Component />);
+  });
+
+  it('should match snapshot', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render a `<AdapterContext.Consumer>`', () => {
```
tdeekens: Memo: test prop passing through Provider.

packages/react/modules/components/reconfigure-adapter/reconfigure-adapter.js
```diff
@@ -0,0 +1,53 @@
...
+
+export class ReconfigureAdapter extends PureComponent<Props> {
+  static displayName = 'ReconfigureAdapter';
+
+  static defaultProps = {
+    exact: false,
+    children: null,
+  };
+
+  componentDidMount(): Promise<any> | void {
```
tdeekens: Memo: wrong return. It is always void.

readme.md
```diff
@@ -147,6 +147,8 @@ without Redux.
 
 * `ConfigureFlopFlip` a component to configure flopflip with an adapter
   (alternative to the store enhancer)
+* `ReconfigureFlopFlip` a component to reconfigure flopflip with new user properties
```
tdeekens: Memo: share reasoning and illustrate example below.

tdeekens:
Tested this with two `ReconfigureFlopflip`s in the `demo` and a patched `localstorage-adapter` which delays initialization by 500ms. The different `user` props are merged into `adapterArgs` and a call to `reconfigure` on the adapter is called.

packages/react/modules/components/configure-adapter/configure-adapter.js
```diff
@@ -27,19 +23,96 @@ type Props = {
...
 
   adapterState: AdapterState = AdapterStates.UNCONFIGURED;
-  setAdapterState = (nextAdapterState: AdapterState): void => {
+  pendingAdapterArgs: ?AdapterArgs = null;
+
+  state: { adapterArgs: AdapterArgs } = {
+    adapterArgs: this.props.adapterArgs,
+  };
+
+  setAdapterState = (
```
emmenko: Nitpick: isn't this meant to be more like a `cacheAdapterState`, since it's stored in the component instance?

packages/react/modules/components/configure-adapter/configure-adapter.js
```diff
@@ -27,19 +23,96 @@ type Props = {
...
     children: null,
     defaultFlags: {},
   };
 
   adapterState: AdapterState = AdapterStates.UNCONFIGURED;
-  setAdapterState = (nextAdapterState: AdapterState): void => {
+  pendingAdapterArgs: ?AdapterArgs = null;
+
+  state: { adapterArgs: AdapterArgs } = {
+    adapterArgs: this.props.adapterArgs,
```
emmenko: Question: as far as I can see, the `props.adapterArgs` is only used to "initialize" the state.
Wouldn't be better to name it like `props.defaultAdapterArgs`?
It might be less confusing compared to `state.adapterArgs`.

Also, the `ConfigureAdapter` does not check if the `props.adapterArgs` changed, thus updating its internal state. This would be now the job of the `ReconfigureAdapter` I suppose.
Just for me to clarify this point.

packages/react/modules/components/configure-adapter/configure-adapter.js
```diff
@@ -27,19 +23,96 @@ type Props = {
...
+
+  /**
+   * NOTE:
+   *   This is passed through the React context (it's a public API).
+   *   Internally this component has a `ReconfigureAdapter` type;
+   *   this function has two arguments for clarify.
+   */
+  reconfigureOrQueue = (
+    nextAdapterArgs: AdapterArgs,
+    options: AdapterReconfigurationOptions = {}
```
emmenko: Nitpick: you can have the default value in the `mergeAdapterArgs`, `reconfigureOrQueue` does not use `options`.

packages/react/modules/components/configure-adapter/configure-adapter.js
```diff
@@ -27,19 +23,96 @@ type Props = {
...
+  setPendingAdapterArgs = (
+    nextReconfiguration: AdapterReconfiguration
+  ): void => {
+    /**
+     * NOTE:
+     *    The next reconfiguration is merged into the previous
+     *    one instead of maintaining a queue.
+     */
+    this.pendingAdapterArgs = mergeAdapterArgs(
+      this.pendingAdapterArgs || this.state.adapterArgs,
```
emmenko: Hmm, can you clarify this choice?

packages/react/modules/components/configure-adapter/configure-adapter.spec.js
```diff
@@ -75,10 +82,11 @@ describe('lifecycle', () => {
...
               </ConfigureAdapter>
             );
-          });
 
-          beforeEach(() => {
-            wrapper.instance().componentDidMount();
+            wrapper
+              .instance()
+              .componentDidMount()
+              .catch(() => {});
```
emmenko: Hmm, shouldn't you do `return wrapper...`? `componentDidMount` returns a Promise, so `beforeEach` should wait for it to be resolved I suppose, or?

packages/react/modules/components/configure-adapter/configure-adapter.js
```diff
@@ -27,19 +23,96 @@ type Props = {
...
+export const AdapterContext: Context<ReconfigureAdapter> = createReactContext(
+  () => {}
+);
 
-export default class ConfigureAdapter extends PureComponent<Props> {
+export const mergeAdapterArgs = (
+  previousAdapterArgs: AdapterArgs,
+  { adapterArgs: nextAdapterArgs, options }: AdapterReconfiguration
+): AdapterArgs =>
+  options.exact ? nextAdapterArgs : merge(previousAdapterArgs, nextAdapterArgs);
```
emmenko: I would suggest to rename this prop `exact` to something like `shouldMergeAdapterArgs`.
For me it's a bit confusing to understand what this prop means and does.
When I see I also think about the `<Route exact/>`, which in that case means "to match the exact path", but in this case it's used as "should merge the props or override them".

packages/react/modules/components/configure-adapter/configure-adapter.spec.js
```diff
@@ -328,10 +360,269 @@ describe('lifecycle', () => {
...
+        wrapper.instance().setPendingAdapterArgs({
+          adapterArgs: nextAdapterArgs,
+          options: { exact: false },
+        });
+
+        wrapper.instance().unsetPendingAdapterArgs();
+      });
+
+      it('should invoke `setAdapterArgs`', () => {
+        expect(wrapper.instance().setAdapterArgs).toHaveBeenCalled();
```
emmenko: Nitpick: I usually avoid this test if I also have `toHaveBeenCalledWith`, because the latter implies the function "to have been called".

packages/react/modules/components/configure-adapter/with-reconfiguration.js
```diff
@@ -0,0 +1,33 @@
...
+) => {
+  class EnhancedComponent extends PureComponent<
+    $Diff<RequiredProps, ProvidedProps>
+  > {
+    static displayName = wrapDisplayName(Component, 'withReconfiguration');
+    render(): Node {
+      return (
+        <AdapterContext.Consumer>
+          {reconfigure => (
+            <Component {...{ [propKey]: reconfigure }} {...this.props} />
```
emmenko: Shouldn't it be the other way around? Otherwise `...this.props` might accidentally override `[propKey]`

packages/react/modules/components/configure-adapter/with-reconfiguration.spec.js
```diff
@@ -0,0 +1,35 @@
...
+import withReconfiguration from './with-reconfiguration';
+
+const TestComponent = () => <div />;
+TestComponent.displayName = 'TestComponent';
+
+describe('rendering', () => {
+  let wrapper;
+
+  beforeEach(() => {
+    const Component = withReconfiguration('foo-prop-key')(TestComponent);
```
emmenko: You probably want to test it also without the custom prop name. Also to check that the wrapped component gets the `reconfigure` prop?

packages/react/modules/components/reconfigure-adapter/reconfigure-adapter.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  render(): Node {
+    return this.props.children
+      ? React.Children.only(this.props.children)
+      : null;
+  }
+}
+
+export default withReconfiguration()(ReconfigureAdapter);
```
emmenko: Hmm, wouldn't it make more sense to have the react component using the default implementation and the HOC reusing the react component?
Additionally, I would simply export the HOC from this module as a named export instead of a separate file.

readme.md
```diff
@@ -249,6 +251,38 @@ import adapter from '@flopflip/launchdarkly-adapter';
 </ConfigureFlopFlip>;
 ```
 
+Whenever your application "gains" certain information (e.g. with `react-router`) only further
+down the tree but that information should be used for user targeting (through `adapterArgs.user`) you
+can use `ReconfigureFlopflip`. `ReconfigureFlopflip` itself communicates with `ConfigureFlopflip`
+to reconfigure the given adapter for more fine grained targeting with the pased `user`.
```
emmenko: ~~pased~~ `passed`

readme.md
```diff
@@ -249,6 +251,38 @@ import adapter from '@flopflip/launchdarkly-adapter';
...
+You also do not have to worry about rendering any number of `ReconfigureFlopflip`s before the adapter is
+initialized (e.g. LaunchDarkly). Requested reconfigurations will be queued and processed once the adapter is ready.
+
+Imagine having `ConfigureFlopflip` above a given component wrapped by a `Route`:
```
emmenko: For completeness I would include `ConfigureFlopFlip` in the code example below

readme.md
```diff
@@ -249,6 +251,38 @@ import adapter from '@flopflip/launchdarkly-adapter';
...
+```jsx
+<Route
+  exact={false}
+  path="/:projectKey"
+  render={routerProps => (
+    <React.Fragment>
+      <MyRouteComponent />
+      <ReconfigureFlopflip
+        exact={false}
+        // Note: this should be memoised to not trigger wasteful `reconfiguration`s.
```
emmenko: Good point! If this is important, you should mention this more clearly in the documentation (non only in the example). Is it possible maybe to do this in the internal implementation, so the user does not have to worry about it?

readme.md
```diff
@@ -249,6 +251,38 @@ import adapter from '@flopflip/launchdarkly-adapter';
...
+        user={{ projectKey: routerProps.projectKey }}
+      />
+    </React.Fragment>
+  )}
+/>
+```
+
+The passed `projectKey` will be passed to the adapter from `ReconfigureFlopflip` up to `ConfigureFlopflip` automatically
```
emmenko: Nitpick:

> Internally, `ReconfigureFlopFlip` will pass the `projectKey` to `ConfigureFlopFlip`, causing the adapter to automatically update the user context and therefore to flush the adapter flags.

tdeekens:
Thanks for the feedback. I'll incorporate it and maybe release it as a fix.

packages/react/modules/components/configure-adapter/configure-adapter.js
tdeekens: Not sure. Maybe it's more `appliedAdapterArgs` as soon as they end up on component state they should be applied to the adapter. Will rename.

packages/react/modules/components/configure-adapter/configure-adapter.js
tdeekens: Yeah that's actually true. I think people would expect the adapterArgs to update when the props change. Which is actually also part of the crux of being able to change them from the top and from below.

packages/react/modules/components/configure-adapter/configure-adapter.js
tdeekens: Sure. Imagine you have two `ReconfigureFlopflip` components somewhere flying around you component tree. One with `user={{ firstRun: 'foo' }}` the second with `user={{ secondRun: 'bar' }}`. Both may render (trigger reconfiguration) before the adapter even initialized. Given that the first queues a reconfiguration the `pendingAdapterArgs` are populated. Given that after the second ran we want to reconfigure with merged `adapterArgs` of both we have to `mergeAdapterArgs` with the `pendingsAdapterArgs` as using the `state.appliedAdapterArgs` would imply the that lastly rendered `ReconfigureFlopflip` always wins. Hope that makes sense.

packages/react/modules/components/configure-adapter/configure-adapter.spec.js
tdeekens: True. I guess it just happened to work.

packages/react/modules/components/configure-adapter/with-reconfiguration.js
tdeekens: Yeah I can change that true.

packages/react/modules/components/reconfigure-adapter/reconfigure-adapter.js
tdeekens: Thought about the latter too but liked working with the separate file. I kind of like these short files/components to not have to scroll around too much.

For the first: yeah, I wanted to always have a render-prop based version and the HoC using that. In this case the HoC only wraps the `Context.Consumer` to I didn't feel too much value in extracting that yet. Hope that makes sense.

readme.md
tdeekens: Hm, not sure about the implementation but. I guess one would have to deep equal the user prop. Not sure I want that as I rather would like to "encourage" best practices somewhat.

packages/react/modules/components/configure-adapter/configure-adapter.js
emmenko: Hmm thanks for the explanation. I’m still a bit confused though.
For example, can’t we check if pendingAdapterArgs is defined and only then do the merge with the state.appliedAdapterArgs?

Also, what would happen if:
* first reconfigure renders (Adapter not ready)
* Adapter ready
* Second reconfigure renders

tdeekens: I think I've a knot in my brain. We have to merge

1. With `pendingAdapterArgs` whenever multiple calls to `reconfigureOrQueue` are performed while the adapter was still configuring. To accumulate the `pendingAdapterArgs` (merged instead of maintaining a queue).
2. With `appliedAdapterArgs` whenever the adapter is configured and is able to "flush" the `nextAdapterArgs` straight away.

The first time round we merge with `appliedAdapterArgs` (from `componentDidMount`) as `nextAdapterArgs` passed to `reconfigureOrQueue` might be only "shallow".

To you example:

1. Adapter not ready: `reconfigureOrQueue` is called -> `pendingAdapterArgs` are a deep merge of `appliedAdapterArgs` and the passed `nextAdapterArgs`
2. Adapter ready: the `pendingAdapterArgs` are applied through `unsetPendingAdapterArgs` and "flushed" through `setAdapterArgs` (which could be renamed to `applyAdapterArgs`
3. Adapter still ready: Second call to `reconfigureOrQueue` -> No call to `setPendingAdapterArgs`, straight to `setAdapterArgs` which flushes them straight away.

Hope I make sense. Please correct me if I am wrong.

PR #285, opened 2018-04-01
Improve `ReconfigureFlopflip`

tdeekens:
This applies some of feedback provided by @emmenko. Thanks!

https://github.com/tdeekens/flopflip/pull/280#discussion_r178456833

Maybe you can glimpse over to see if I got things right.


tdeekens:
As a note. I am thinking of renaming `adapterArgs` on `ConfigureFlopflip` to `defaultAdapterArgs` and only allow post-render mutation of them via `ReconfigureFlopflip`. It would obviously be a breaking change though and I've received feedback on issues indicating that users quite like the controlled nature and auto-reconfiguring through `ConfigureFlopflip`.

PR #311, opened 2018-04-08
Configure Fossa for license compliance checking

tdeekens:


tdeekens:
That service is promising but flaky. I can't locally install their CLI.

PR #319, opened 2018-04-11
chore: update create-react-context

tdeekens:
> Should be careful with this. An update broke in a previous version without reason.

Will link this into an existing project to ensure that it works.


tdeekens:
This error in unit tests it also what shows in during integration in an application.

Issue #324, opened 2018-04-13
`shouldDeferAdapterConfiguration` not working as intended


Kerumen:
I have an issue with `shouldDeferAdapterConfiguration` in my app. I'll try to explain it: I have an XHR call which populate a user. So I set `shouldDeferAdapterConfiguration={!user.key}` and flopflip should not initialize until my user is ready. This is not the case. I investigated and maybe I found why.

First the component `ConfigureFlopFlip` mount and this condition match so the adapter is not configured. Good behaviour.
https://github.com/tdeekens/flopflip/blob/60d023c7c42efad54d9d0249dabb89d16cf1289f/packages/react/modules/components/configure-adapter/configure-adapter.js#L140

Then, when my user is ready, `componentWillReceiveProps` is called with `this.reconfigureOrQueue`.
https://github.com/tdeekens/flopflip/blob/60d023c7c42efad54d9d0249dabb89d16cf1289f/packages/react/modules/components/configure-adapter/configure-adapter.js#L133

The adapter is not yet configured so `setPendingAdapterArgs` is called
https://github.com/tdeekens/flopflip/blob/60d023c7c42efad54d9d0249dabb89d16cf1289f/packages/react/modules/components/configure-adapter/configure-adapter.js#L87

My new user lives now in `this.pendingAdapterArgs` ready to be passed to the adapter but `applyPendingAdapterArgs` is never called
https://github.com/tdeekens/flopflip/blob/60d023c7c42efad54d9d0249dabb89d16cf1289f/packages/react/modules/components/configure-adapter/configure-adapter.js#L105

`componentDidUpdate` occurs and now, `this.props.shouldDeferAdapterConfiguration` is `false` so `configure` is called, however `this.state.appliedAdapterArgs` doesn't have my user. So an error is triggered LaunchDarkly side.
https://github.com/tdeekens/flopflip/blob/60d023c7c42efad54d9d0249dabb89d16cf1289f/packages/react/modules/components/configure-adapter/configure-adapter.js#L166

I don't have enough visibility on the code yet to provide a fix. Let me know if any information can help :)

tdeekens:
Thanks for the detailed description. I'll try to have a look this weekend. Sorry for the inconvenience!

tdeekens:
[...]

PR #329, opened 2018-04-14
Fix `ConfigureAdapter` with deferred configuration

tdeekens:
> The pull request closes an issue found by @Kerumen which made the adapter configure with false `adapterArgs` when deferring its configuration.

The issue #324 describes the problem more eloquently than I could.

One plausible solution is outlined in this PR. The adapter has to configure or reconfigure with queued (pending) or applied (from initial state) args. This situation is a bit "tricky" as we can not call `setState` in `componentDidUpdate` and also do not want the adapter be configured first with stale args and then with up-to-date ones.

### DIFF
--- a/packages/react/modules/components/configure-adapter/configure-adapter.js
+++ b/packages/react/modules/components/configure-adapter/configure-adapter.js
@@ -101,16 +101,33 @@ export default class ConfigureAdapter extends PureComponent<Props, State> {
      * NOTE:
      *    The next reconfiguration is merged into the previous
      *    one instead of maintaining a queue.
+     *
+     *    The first merge with merge with `appliedAdapter` args
+     *    to contain the initial state (through property initializer).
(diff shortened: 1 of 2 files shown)


tdeekens:
@Kerumen would you mind having a look just to make sure things make sense and I didn't miss anything?

tdeekens:
Thanks. Yeah that's what I am trying to also still wrap my head around. Essentially, I want to be careful with `componentDidUpdate` as `appliedAdapterArgs` is on the state I added the `pendingAdapterArgs` not to be. So they can be queued up e.g. by a deferred configuration or when rendering `ReconfigureFlopflip` components before the adapter is ready.

Kerumen:
@tdeekens It works well! Thanks :)

tdeekens:
Let's release it then.

PR #332, opened 2018-04-15
Fix yarn caching on CircleCI

tdeekens:


.circleci/config.yml
```diff
@@ -4,14 +4,14 @@ aliases:
...
-      - v2-yarn-cache
+      - v3-yarn-cache-{{ checksum "yarn.lock" }}
+      - v3-yarn-cache
 
   - &save_yarn_cache
     name: Saving cache
-    key: v2-yarn-cache-{{ checksum "yarn.lock" }}
+    key: v3-yarn-cache-{{ checksum "yarn.lock" }}
     paths:
-      - node_modules
```
emmenko: NOTE that now you always need to run `yarn install` after you've restored the cache in order to recreate `node_modules`.
I think in this repo it's fine and you probably don't need to do that. 

Just a heads-up in case you "only" restore the cache 😉

tdeekens: Yeah that's true. But I find it rather safe and logical to do that. Just hoping that copying over some folder will make things work is quite optimistic 😄.

PR #363, opened 2018-04-27
fix(react/configure): unsetting of pending adapter args

tdeekens:


packages/react/modules/components/configure-adapter/configure-adapter.js
```diff
@@ -69,10 +69,20 @@ export default class ConfigureAdapter extends PureComponent<Props, State> {
...
+       *   a batched `setState` otherwise outdated `adapterArgs` may be
+       *   applied as the `setState` is actually performed later.
+       */
+      this.pendingAdapterArgs = null;
+
+      return {
+        ...prevState,
+        appliedAdapterArgs: nextAdapterArgs,
+      };
+    });
```
dferber90: I don't fully get the comment... but I'm missing a ton of context anyways.

Would one of these two solutions be preferable?

```js
this.pendingAdapterArgs = null;
this.setState(prevState => {})
```

```js
this.setState(
  prevState => {},
  () => { this.pendingAdapterArgs = null; }
)
```

Just a suggestion as I've never seen a side-effect within the `setState`.

tdeekens: Oh shit. Thanks. Will do

packages/react/modules/components/configure-adapter/configure-adapter.spec.js
```diff
@@ -444,7 +464,7 @@ describe('interacting', () => {
       beforeEach(() => {
         wrapper.instance().setPendingAdapterArgs({
           adapterArgs: nextAdapterArgs,
-          options: { exashouldOverwritect: false },
```
emmenko: LoL 😅

tdeekens: Yeah late night coding doesn't do anyone good :)

PR #441, opened 2018-08-02
TypeScript: correct casing of export ConfigureFlopFlip

omichelsen:
The exported function is called `ConfigureFlopFlip` and not `ConfigureFlopflip`.


tdeekens:
Will release this weekend.

Issue #475, opened 2018-09-04
Allow reset of memory-adapter


dferber90:
First of all: Lovely package :)

I'm using the `memory-adapter` in an integration test to provide the default flags like this:

```jsx
import memoryAdapter from '@flopflip/memory-adapter'

<ConfigureFlopFlip
  adapter={memoryAdapter}
  defaultFlags={{ someFeature: true }}
>
  <ComponentUnderTestWhichUsesFlags />
</ConfigureFlopFlip>
```

It would be great if it was possible to reset the `memoryAdapter` after each test like so

```js
afterEach(() => {
  memoryAdapter.reset()
})
```

Do you think this is a good idea?

The closest thing to this I've found would be to do `memoryAdapter.reconfigure()`. Or should we have a dedicated `test-adapter`?

tdeekens:
Reset on an adapter would be good. Could also work for all adapters as I‘d like all to have the samd API. The only concern is that the defaultFlags would go onto the adaperState and serve as the basis for a reset. Nothing too bad just how it could work I suppose. 

This reset would mostly be for testing, right? I‘ve heard people also demanding a locked-in adapter. Essentially not updating it‘s flags after it received them initially. Not related but something which I‘d like to maybe to in one release.

tdeekens:
I think you would need to have an onChange on the adapter to trigger the change handlers passed by ConfigureAdapter. You expose reset on the adapter which internally onChanges to the defaultFlags. Might need some shifting of code to make it easier to follow.

dferber90:
Thanks for going forward with this! I'm looking at the feature request purely from a consumer perspective wanting to write tests.

An alternative to having a `reset` method would be to turn the adapter into a factory. Then we wouldn't need the reset at all.

Then the tests could be:

```js
import createMemoryAdapter from '@flopflip/memory-adapter'

describe('ComponentUnderTestWhichUsesFlags', () => {
  it('should render "hello" when someFeature is on', () => {
    const memoryAdapter = createMemoryAdapter()
    const { getByText } = render(
      <ConfigureFlopFlip
        adapter={memoryAdapter}
        defaultFlags={{ someFeature: true }}
      >
        <ComponentUnderTestWhichUsesFlags />
      </ConfigureFlopFlip>
    )
    expect(getByText('Hello')).toBeVisible();
  })
  it('should not render "hello" when someFeature is off', () => {
    const memoryAdapter = createMemoryAdapter()
    const { queryByText } = render(
[...]

PR #476, opened 2018-09-04
Add `reset` to `memory-adapter`

tdeekens:
Closes #476 

This adds the ability to reset the memory-adapter as discussed in #475. I started 
this to have a discussion around it.


packages/memory-adapter/modules/adapter/adapter.js
```diff
@@ -16,6 +16,7 @@ const adapterState: AdapterState = {
 
 const configure = ({
   user,
+  defaultFlags = {},
```
tdeekens: The adapter receives the defaultFlags (named this way if I'd start again I'd name it initialFlags.

packages/memory-adapter/modules/adapter/adapter.js
```diff
@@ -36,12 +38,17 @@ const configure = ({
 const reconfigure = ({ user }: { user: User }): Promise<any> => {
   updateUser(user);
 
-  adapterState.flags = {};
-  adapterState.onFlagsStateChange({});
+  adapterState.flags = adapterState.defaultFlags;
+  adapterState.onFlagsStateChange(adapterState.defaultFlags);
```
tdeekens: You reset to the defaultFlags not the empty thing.

tdeekens:
I am not sure if this should really "spread" to other adapters. It feels like calling `ldAdapter.reset()` gives the wrong impression that something also happens upstream (potentially). In that I kind of like the simplicity of this now. As testing purposes are your use case it would work for you, correct?

tdeekens:
I want to kill myself right now. One one machine snapshots pass and don't give updates. On CI they fail. On another machine they fail and give updates. Committing them makes CI still fail.

PR #477, opened 2018-09-04
Add `children` (FaaC) and render prop support to `ConfigureAdapter`

### LINKED ISSUE #473: Loading state not exposed to app
It would be nice if a `loading` prop was available either as a function as child argument, so that consumers could opt in to not rendering their apps until the connection to launch darkly has succeeded (or failed)

tdeekens:
Closes #473 

Please refer to the readme for examples. Essentially it adds exposing ready state to all adapters while exposes it through `ConfigureAdapter`.


readme.md
```diff
@@ -255,6 +255,24 @@ import adapter from '@flopflip/launchdarkly-adapter';
 </ConfigureFlopFlip>;
 ```
 
+You can also pass `render` or `children` as a function to act differently based on the underlying adapter's ready state:
+
+```jsx
+<ConfigureFlopFlip adapter={adapter} adapterArgs={{ clientSideId, user }}>
+  {{isAdapterReady} => isAdapterReady ? <App /> : <LoadingSpinner />}
```
tdeekens: This is not 100% recommended as your app should work without flags and not rely on an external service for boot.

tdeekens:
@AndrewRayCode please verify that this serves your use case.

PR #479, opened 2018-09-05
Add ability to opt-out of flag subscription of ld-adapter

### LINKED ISSUE #474: Add option to not live update from feature flagging upstream
We have some mission critical workflows where we don't want the UI to live shift out from under our users. It would be nice to have the option to disable a live subscription to the flagging system.

I could probably hack around this with a HOC wrapper holding a `<ConfigureFlopFlip />` that only passes the initial flag load down to components, but it would be nice to have this first class.

tdeekens:
Closes #474 

This adds the ability to opt-out of the flags subscription by passing a `shouldSubcribeToFlagChanges: false` in the `adapterArgs` passed to `ConfigureFlopflip`.

### DIFF
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.js
@@ -35,7 +35,7 @@ const adapterState: AdapterState = {
   flags: null,
 };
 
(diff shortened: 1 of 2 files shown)


AndrewRayCode:
Looks great to me! I think naming the prop just `subscribeToFlagsChanges` would make sense too

PR #502, opened 2018-09-28
docs(readme): update title to `why you might use this`

adnasa:
I like this wording better ❤️


tdeekens:
...so true. I am so glad you changed this!

PR #507, opened 2018-10-27
Refactor bundling to exclude deps when not building umd

tdeekens:
#### Summary

This pull request changes bundling to exclude any dependency (mark it as external) when not bundling umd.

#### Description

Given flopflip is bundled again by a module bundler (which it has to when running cjs or umd) all deps will be bundled again anyway. Given this I doubt that tree shaking can be effective and suspect double dependency bundling.

This is something @montezume and I discovered recently and aligns with kcd-scripts.

Furthermore, I replaced uglify with terser. I will likely publish this as a canary release first to test it in app.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,17 +1,38 @@
+const path = require('path');
 const resolve = require('rollup-plugin-node-resolve');
 const commonjs = require('rollup-plugin-commonjs');
 const babel = require('rollup-plugin-babel');
 const replace = require('rollup-plugin-replace');
-const { uglify } = require('rollup-plugin-uglify');
+const { terser } = require('rollup-plugin-terser');
 const json = require('rollup-plugin-json');
 const builtins = require('rollup-plugin-node-builtins');
 const globals = require('rollup-plugin-node-globals');
 const flow = require('rollup-plugin-flow');
 const filesize = require('rollup-plugin-filesize');
 const babelOptions = require('./babel.config');
+const pkg = require(path.join(process.env.INIT_CWD, 'package.json'));
 
 const env = process.env.NODE_ENV;
 const name = process.env.npm_package_name;
+const format = process.env.npm_lifecycle_event.split(':')[1];
+
+const pkgDependencies = Object.keys(pkg.dependencies || {});
+const pkgPeerDependencies = Object.keys(pkg.peerDependencies || {});
+const pkgOptionalDependencies = Object.keys(pkg.optionalDependencies || {});
+
+/**
+ * Note:
+ *   Given we do not bundle for UMD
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -1,17 +1,38 @@
...
 const builtins = require('rollup-plugin-node-builtins');
 const globals = require('rollup-plugin-node-globals');
 const flow = require('rollup-plugin-flow');
 const filesize = require('rollup-plugin-filesize');
 const babelOptions = require('./babel.config');
+const pkg = require(path.join(process.env.INIT_CWD, 'package.json'));
 
 const env = process.env.NODE_ENV;
 const name = process.env.npm_package_name;
+const format = process.env.npm_lifecycle_event.split(':')[1];
```
tdeekens: Sadly rollup does not expose its own CLI option again on the `process.env`. This is a kind of CoC approach to extract the format from the lifecycle.

tdeekens:
Checking the build output: umd now contains all deps, all other targets don't.

rollup.config.js
```diff
@@ -1,17 +1,38 @@
...
 const replace = require('rollup-plugin-replace');
-const { uglify } = require('rollup-plugin-uglify');
+const { terser } = require('rollup-plugin-terser');
 const json = require('rollup-plugin-json');
 const builtins = require('rollup-plugin-node-builtins');
 const globals = require('rollup-plugin-node-globals');
 const flow = require('rollup-plugin-flow');
 const filesize = require('rollup-plugin-filesize');
 const babelOptions = require('./babel.config');
+const pkg = require(path.join(process.env.INIT_CWD, 'package.json'));
```
montezume: how about using a package like [read-pkg-up](https://github.com/sindresorhus/read-pkg-up)?

tdeekens: Thanks 💯. Just when you think Sindre didn't make a package for it.

montezume: @tdeekens just found it a few days ago in kcd-scripts. 😆

tdeekens: Tried, the promise virus spreads. Top-level async/await is not a thing in Node.js yet. Making things a bit odd. I'd stick with this for now but thanks for the suggestion!

tdeekens: Wait a minute: `readPkgUp.sync`.

montezume: Yeah, that's what I was going to suggest 🤷‍♂️

montezume: [here](https://github.com/kentcdodds/kcd-scripts/blob/8337bbd16fa4f964f688975f3c510d4705c94932/src/utils.js#L10)

PR #511, opened 2018-11-04
feat(react-broadcast): add useFeatureToggle react hook

tdeekens:
#### Summary

This pull request adds a `useFeatureToggle(flagName: FlagName, flagVariation: FlagVariation)` React hook.

#### Description

```js
import { useFeatureToggle } from '@flopflip/react-broadcast';

function ComponentWithFeatureToggle(props) {
   const isFeatureEnabled = useFeatureToggle('myFeatureToggle');

   return (
     <h3>{props.title}<h3>
     <p>
       The feature is {isFeatureEnabled ? 'enabled' : 'disabled'}
     </p>
   );
}
```

This is just a small experiment how we flopflip's architecture could support a React hook. It is unclear to me how this could also be supported for `@flopflip/react-redux` without redux having a `useRedux` hook to use.

### DIFF
--- a/packages/react-broadcast/modules/hooks/index.js
+++ b/packages/react-broadcast/modules/hooks/index.js
@@ -1,0 +1,1 @@
+export { default as useFeatureToggle } from './use-feature-toggle';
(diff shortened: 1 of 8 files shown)


tdeekens:
Will close again as this will not be part of the library until hooks have landed in React itself.

PR #516, opened 2018-11-10
Remove `recompose`

tdeekens:
#### Summary

This pull request removes recompose and replaces it with small and custom hocs.

#### Description

Recompose will likely be discontinued while flopflip at most used 20% of it. Tree shaking might have helped but those 2-3 hocs can be replaced by custom built ones. 

I really thought this would be easier but it felt somewhat like taking out really essential glue code. I feel that this will be even harder to application level code.

### DIFF
--- a/packages/react/modules/hocs/with-props/with-props.js
+++ b/packages/react/modules/hocs/with-props/with-props.js
@@ -1,5 +1,6 @@
 import React, { createElement } from 'react';
 import { wrapDisplayName } from '../wrap-display-name';
+import { setDisplayName } from '../set-display-name';
 
 const withProps = mapProps => BaseComponent => {
   const WithProps = ownProps => {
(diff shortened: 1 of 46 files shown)


packages/react/modules/hocs/with-props/with-props.js
```diff
@@ -0,0 +1,31 @@
...
+import { wrapDisplayName } from '../wrap-display-name';
+import { setDisplayName } from '../set-display-name';
+
+type WithProps = {};
+type ProvidedProps = {};
+
+const withProps = (
+  mapProps: WithProps | ((ownProps: ProvidedProps) => WithProps)
+) => (BaseComponent: ComponentType<ProvidedProps & WithProps>) => {
+  const WithProps = (ownProps: ProvidedProps): ProvidedProps & WithProps => {
```
tdeekens: Typing this was fun.

PR #520, opened 2018-11-12
Add `react-is` to prop-types (suspense ready)

tdeekens:
#### Summary

react-is has to be used to evaluate anything as a React component with suspense. As e.g. anything returned from `React.lazy(() => import(...))` will otherwise fail prop-types validation.


packages/react/modules/components/toggle-feature/toggle-feature.js
```diff
@@ -31,6 +33,25 @@ export default class ToggleFeature extends PureComponent<Props> {
     children: null,
   };
 
+  static propTypes = {
+    untoggledComponent: (props: Props, propName: string) => {
+      if (props[propName] && !isValidElementType(props[propName])) {
+        return new Error(
+          `Invalid prop 'untoggledComponent' supplied to 'Route': the prop is not a valid React component`
+        );
```
tdeekens: This is how react-router also validates it.

packages/react/modules/components/reconfigure-adapter/reconfigure-adapter.js
```diff
@@ -24,6 +25,11 @@ export class ReconfigureAdapter extends PureComponent<Props> {
     children: null,
   };
 
+  static propTypes = {
+    shouldOverwrite: PropTypes.bool,
+    children: PropTypes.node,
```
tdeekens: Feels like having these can't hurt. Another option would be to generate them from flow types but that also feels a bit over done for now.

packages/react/modules/components/configure-adapter/configure-adapter.spec.js
```diff
@@ -26,7 +26,7 @@ const createTestProps = props => ({
     onStatusStateChange: jest.fn(),
   },
   adapter: createAdapter(),
-  children: ChildComponent,
+  children: <ChildComponent />,
```
tdeekens: That was actually a "bug" in the test.

PR #634, opened 2019-03-22
update typings to fix build errors

ChrisW-B:
#### Summary

Updated a few types to be more in line with tslint/eslint settings, fixed one type that was causing build issues (typescript assumed a variable's name was being changed to boolean)


tdeekens:
Released in

1. react-broadcast@8.0.3
2. react-redux@8.0.3

PR #681, opened 2019-05-14
refactor(launchdarkly-adapter): to waitForInitialization

### LINKED ISSUE #679: `ready` event is emitted on error when initializing client
**Is your feature request related to a problem? Please describe.**

A `LaunchDarklyFlagFetchError` is sometimes thrown during client initialization. This seems to occur because the `ready` event gets emitted as soon as the initialization logic has finished, regardless of whether it succeeded as explained here: https://github.com/launchdarkly/js-client-sdk/issues/113#issuecomment-422123189

**Describe the solution you'd like**

The `ready` event is only emitted on successful initialization

**Describe alternatives you've considered**

The `waitUntilReady` method call here
https://github.com/tdeekens/flopflip/blob/9537bd28d54f384f53e99907b05e3897bb50ff2e/packages/launchdarkly-adapter/modules/adapter/adapter.ts#L151
could be replaced with the `waitForInitialization` method as discussed here:
https://github.com/launchdarkly/js-client-sdk/issues/113#issuecomment-424922537

tdeekens:
- Closes #679 

#### Summary

This pull request intends to solve the issue with resolving configuration only with ready event when it succeeds.

### DIFF
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
@@ -179,6 +179,57 @@ describe('when configuring', () => {
       });
     });
 
+    describe('when `waitForInitialization` throws', () => {
+      describe('when it should `throwOnInitializationFailure`', () => {
+        beforeEach(() => {
+          onStatusStateChange = jest.fn();
+          onFlagsStateChange = jest.fn();
+          client = createClient({
+            waitForInitialization: jest.fn(() => Promise.reject()),
+          });
+
+          ldClient.initialize.mockReturnValue(client);
+        });
+
+        it('should reject the configuration with an error', async () => {
+          await expect(
+            adapter.configure({
+              clientSideId,
+              user: userWithKey,
+              onStatusStateChange,
+              onFlagsStateChange,
+              throwOnInitializationFailure: true,
+            })
+          ).rejects.toThrow(
+            '@flopflip/launchdarkly-adapter: adapter failed to initialize.'
+          );
+        });
+      });
+      describe('when it should not `throwOnInitializationFailure`', () => {
+        beforeEach(() => {
+          onStatusStateChange = jest.fn();
+          onFlagsStateChange = jest.fn();
+          client = createClient({
+            waitForInitialization: jest.fn(() => Promise.reject()),
+          });
+
+          ldClient.initialize.mockReturnValue(client);
+        });
+
+        it('should resolve the configuration', async () => {
+          await expect(
+            adapter.configure({
+              clientSideId,
+              user: userWithKey,
+              onStatusStateChange,
+              onFlagsStateChange,
+              throwOnInitializationFailure: false,
+            })
+          ).resolves.toEqual(expect.anything());
+        });
+      });
+    });
+
     describe('with flag updates', () => {
       describe('when `subscribeToFlagChanges`', () => {
         beforeEach(() => {
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
@@ -13,7 +13,7 @@ const userWithoutKey = {
 };
 const flags = { 'some-flag-1': true, 'some-flag-2': false };
 const createClient = jest.fn(apiOverwrites => ({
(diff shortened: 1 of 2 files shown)


tdeekens:
@gnerkus I'd need a bit of advise on this: what would be want to happen if the adapter fails to be configured?

1. Just log the fact that it failed and let default flags handle it all
2. Allow users to configure something so their application can react (e.g. in an error boundary

Any thoughts? Thanks.

gnerkus:
`2` would be my preferred choice. It still enables the user perform `1` while allowing for custom error handling.

tdeekens:
Thanks. Yeah I agree it feels like the most flexible approach. That could imply something like passing some options like ` { throwOnInitializationFailure } ` which would easier trigger the error (wrapper or not) being thrown up the tree or being caught and swalloed in the library. Would this make sense?

gnerkus:
Definitely.

tdeekens:
Sorry. I didn't have time to get to this in the evenings. Will get to it this weekend the lastest.

packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
```diff
@@ -179,6 +179,57 @@ describe('when configuring', () => {
...
+          onFlagsStateChange = jest.fn();
+          client = createClient({
+            waitForInitialization: jest.fn(() => Promise.reject()),
+          });
+
+          ldClient.initialize.mockReturnValue(client);
+        });
+
+        it('should reject the configuration with an error', async () => {
+          await expect(
```
tdeekens: Could return the promise but I like this better. Subjective 🤷‍♂ 😄

packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
```diff
@@ -13,7 +13,7 @@ const userWithoutKey = {
 };
 const flags = { 'some-flag-1': true, 'some-flag-2': false };
 const createClient = jest.fn(apiOverwrites => ({
-  waitUntilReady: jest.fn(() => Promise.resolve()),
+  waitForInitialization: jest.fn(() => Promise.resolve()),
```
tdeekens: ...cause the adapter now uses the fn while defauling into the same behaviour of not rethrowing to have this non-breaking.

packages/launchdarkly-adapter/modules/adapter/adapter.ts
```diff
@@ -142,32 +142,51 @@ export const camelCaseFlags = (rawFlags: Flags): Flags =>
...
+              )
+            );
+          resolve({ flagsFromSdk: null });
+        });
     }
 
-    return reject(new Error('Can not subscribte with non initialized client.'));
+    return reject(
+      new Error(
+        '@flopflip/launchdarkly-adapter: can not subscribe with non initialized client.'
```
gnerkus: Appreciate the error prefixes. 👍

packages/launchdarkly-adapter/modules/adapter/adapter.ts
```diff
@@ -178,13 +197,15 @@ const configure = ({
   onFlagsStateChange,
   onStatusStateChange,
   subscribeToFlagChanges = true,
+  throwOnInitializationFailure = false,
```
gnerkus: I assume this means the default config lets the client fall back to the default flags if initialization fails. So the user decides if they want to throw the error (on initialization failure).

This would be a nice addition to the docs.

gnerkus:
LGTM 👍

packages/launchdarkly-adapter/modules/adapter/adapter.ts
tdeekens: Yes, that is correct. The default behaviour would be, as now, that the default flags kick-in or if none are specific everything defaulting to false. Given the user opts into the "new behaviour" with this option he/she can decide what to do. An React error boundary would catch it. One could also wrap flopflip itself in a custom (more local) error boundary and perform anything the `componentDidCatch`.

tdeekens: Docs would be adequate and help. I would add them after we verified this works. Note also that the `subscribeToFlagChanges` is also a custom property (requested by another user of the library) who found it counter-intuitive that flags can update at any time. Which I still find bizarre cause it's the whole point :)

I would then document both options. Thanks for the reminder.

Issue #688, opened 2019-05-22
React-Redux: typescript types in selectFeatureFlag seem off


jaypea:
**Describe the bug**
ts(2345) Argument of type 'AppState' is not assignable to parameter of type 'Flags'.

**To Reproduce**
given i have a setup like this:
```
import { selectFeatureFlag } from '@flopflip/react-redux';

type AppState  = { 
  ...otherApplicationsStateSlices;
  [FLOPFLIP_STATE_SLICE]: Flags;
};

const mapStateToProps = (state: AppState) => ({
  someOtherState: state.someOtherState,
  isFeatureOn: selectFeatureFlag('fooFlagName')(state),
});

export default connect(mapStateToProps)(FooComponent);
```
Then i get the above mentioned typescript error.

Workaround is ugly and breaking type safety:
`isFeatureOn: selectFeatureFlag('fooFlagName')((state as any) as Flags),`

**Expected behavior**
Have another Type for the state in the definitions instead of :
`export declare const selectFlag: (flagName: string) => (state: Flags) => string | boolean;`
see https://github.com/tdeekens/flopflip/blob/master/packages/react-redux/modules/ducks/flags/flags.ts#L47

Is there another way to use this selector with typescript which i just didn't see?

tdeekens:
Hi,

thanks for reporting! Sorry for the issue with the typing I am personally not using the Redux package so I didn't spot this myself but wanted to provide it for the communicaty. I am not aware of another way of using the selector in a type safe manner. I think calling the `state` flags is false? Maybe just a super set with at least flags?

jaypea:
it may work with just `(state: State)`, similar to https://github.com/tdeekens/flopflip/blob/master/packages/react-redux/modules/ducks/flags/flags.ts#L44

tdeekens:
Thanks. Changed that and published a canary (automatically from `master`). You can try the build with the version: `8.1.3-next.5` in your package.json. Let me know if this works for you. Thanks.

jaypea:
unfortunately this didn't work
type 'AppState' has no properties in common with type 'State'.ts(2559)

my guess now is, it would need an exposed type with the [FLOPFLIP_STATE_SLICE]: flopflipReducer, somehow involved.

tdeekens:
Yeap. Ok, will have a look later when I find time. If you have a clue already: a PR would be welcome :)

jaypea:
the fix worked, thanks for the fast response

[...]

PR #691, opened 2019-05-23
fix(react-redux): state types in react-redux

jaypea:
#### Summary

fix for #688


tdeekens:
Availble as `@flopflip/react-redux@8.1.3-next.6` for testing. Thanks.

PR #692, opened 2019-05-23
fix: splitio-adapter reconfigure

### LINKED ISSUE #689: splitio-adapter: reconfigure does not work
**Describe the bug**
I was trying to configure flopflip with react-redux and  splitio-adapter in the described way.
1. use `createFlopFlipEnhancer` when creating the store. although there is no user-context available during store creation in my app. 
2. during app bootstrap or after login: call reconfigure on the adapter with the users key and details

the reconfigure method in https://github.com/tdeekens/flopflip/blob/master/packages/splitio-adapter/modules/adapter/adapter.ts#L192
does not change the user key and user config in the splitio client objects. 

The splitio client itself does not provide an option to change the user key or user details besides configuring a new client.

this is what i do as a workaround, call `configure` instead of `reconfigure` in my redux actions after login. downside of this is, that i have to maintain some extra state to determine if the user has changed and i cannot use the `createFlopFlipEnhancer`

jaypea:
when calling reconfigure, destroy split.io client and recreate it with saved configuration
fixes #689

### DIFF
--- a/packages/splitio-adapter/modules/adapter/adapter.spec.js
+++ b/packages/splitio-adapter/modules/adapter/adapter.spec.js
@@ -44,9 +44,9 @@ describe('when configuring', () => {
 
   describe('when reconfiguring before configured', () => {
     it('should reject reconfiguration', () => {
-      return expect(
-        adapter.reconfigure({ user: userWithKey, onFlagsStateChange })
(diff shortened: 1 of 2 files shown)


tdeekens:
Looks fine to me. Thanks.

PR #768, opened 2019-08-03
feat(launchdarkly-adapter): adds the ability to `flagsUpdateDelayMs`

### LINKED ISSUE #767: Debounce of trottle adapter updating
Flopflip integrates with vairous adapters. Some of which maintain a websocket connection to the backing services. Anytime these servies change their state and send an event to the respective adapter the app will likely re-render in parts.

This behaviour is obviously intended and would arise with any sort of event based state updates from a remote server. However, an application gives away control over when it is re-rendered. As a result, whenever the flag provider runs into some "update-loop" of needless events being sent the UI will update recurrinly without any need.

As a result we could debounce or throttle the flag update callbacks within the adapters with a given, configured interval. This would also lead to the fact that flags updates are deferred to be only applied e.g. once every 30 seconds.

Thanks @ahmehri for reaching out.

tdeekens:
#### Summary

Closes #767

### DIFF
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
@@ -20,6 +20,11 @@ const createClient = jest.fn(apiOverwrites => ({
   ...apiOverwrites,
 }));
 
+const triggerFlagValueChange = (client, flagValue = false) =>
+  client.on.mock.calls.forEach(([event, cb]) => {
+    if (event.startsWith('change:')) cb(flagValue);
+  });
+
 describe('when configuring', () => {
   let onStatusStateChange;
   let onFlagsStateChange;
@@ -235,10 +240,7 @@ describe('when configuring', () => {
           // Reset due to preivous dispatches
           onFlagsStateChange.mockClear();
 
-          // Checking for change:* callbacks and settings all flags to false.
-          client.on.mock.calls.forEach(([event, cb]) => {
-            if (event.startsWith('change:')) cb(false);
-          });
+          triggerFlagValueChange(client);
         });
 
         it('should `dispatch` `onFlagsStateChange` action', () => {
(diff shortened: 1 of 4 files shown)


packages/launchdarkly-adapter/modules/adapter/adapter.ts
```diff
@@ -77,8 +80,15 @@ const setupFlagSubcription = ({
...
-        onFlagsStateChange(flags);
+        const updateFlags = () => {
+          updateFlagsInAdapterState(flags);
+          onFlagsStateChange(flags);
+        };
+
+        debounce(updateFlags, {
+          wait: flagsUpdateDelayMs,
+          immediate: !flagsUpdateDelayMs,
+        })();
```
tdeekens: This returns a `fn` which you can `.cancel` again but we don't need that. The library's `wait` and `immediate` options allow us to avoid to `if-else` on the argument.

packages/launchdarkly-adapter/modules/adapter/adapter.ts
```diff
@@ -223,6 +235,7 @@ const configure = ({
     if (subscribeToFlagChanges && flagsFromSdk)
       setupFlagSubcription({
         flagsFromSdk,
+        flagsUpdateDelayMs,
```
tdeekens: This would be `<ConfigureFlopflip adapter={ldAdapter} adapterArgs={{ ..., flagsUpdateDelayMs: 1000 }} />

packages/launchdarkly-adapter/modules/adapter/adapter.ts
```diff
@@ -77,8 +80,15 @@ const setupFlagSubcription = ({
...
         };
 
-        updateFlagsInAdapterState(flags);
-        onFlagsStateChange(flags);
+        const updateFlags = () => {
+          updateFlagsInAdapterState(flags);
+          onFlagsStateChange(flags);
+        };
+
+        debounce(updateFlags, {
```
tdeekens: Debounce is on the trailing edge which is what we want I think. Please verify or quickly research so we ensure that.

packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
```diff
@@ -20,6 +20,11 @@ const createClient = jest.fn(apiOverwrites => ({
   ...apiOverwrites,
 }));
 
+const triggerFlagValueChange = (client, flagValue = false) =>
```
tdeekens: Just extracted cause it gets used more often now.

packages/launchdarkly-adapter/modules/adapter/adapter.ts
ahmehri: Verified by reading "Leading edge (or "immediate")" section in [this article](https://css-tricks.com/debouncing-throttling-explained-examples/#article-header-id-1).

Issue #733, opened 2019-07-01
Stricter support for react 16.8 || 16.9


emmenko:
**Is your feature request related to a problem? Please describe.**
Using newer versions of react causes some warnings like

```
Warning: componentWillReceiveProps is deprecated and will be removed in the next major version. Use static getDerivedStateFromProps instead.
    
    Please update the following components: ConfigureAdapter
    
    Learn more about this warning here:
    https://fb.me/react-async-component-lifecycle-hooks
```

**Describe the solution you'd like**
Flopflip should probably drop support for react < 16.8 and rewrite the internal components to drop the deprecated lifecycle methods.

**Describe alternatives you've considered**
Not that I know of, maybe you have other solutions?

**Additional context**

https://github.com/commercetools/merchant-center-application-kit/pull/801#issuecomment-507158312

tdeekens:
> Flopflip should probably drop support for react < 16.8 and rewrite the internal components to drop the deprecated lifecycle methods.

Yes, it's been planned but we want to deprecate only if really needed. So far, apart from being modern which inofitself has no purpose, we didn't plan to drop support for older React versions yet.

tdeekens:
I will close this and plan this as a house keeping milestone. Thanks for opening this and pushing for it :)

PR #774, opened 2019-08-07
chore: support only react >=16.3, rename deprecated lifecycle method

emmenko:
Ref #733


emmenko:
Yeah I suppose it makes sense to split them

PR #842, opened 2019-10-17
feat(launchdarkly-adapter): to allow passing flags

tdeekens:
#### Summary

This pull request allows passing `requestedFlags` to the `launchdarkly-adpter`.

#### Description

Passing `requestedFlags` will change the flag setup so that `variation` is called over `allFlags` which allow using LDs long running flag detection.


tdeekens:
I need to think about adpater reconfiguration:

- If you configure with `{ flagA: false, flagB: true }`
- Then decide to reconfigure with `{ flagA: false, flagC: false }` what's supposed to happen

Additionally tests are missing.

I would value a quick look and feedback towards the above by @thomassvensen :). I named it `requestedFlags` feel free to suggest another name :).

tdeekens:
Will deploy a canary of this from `master`. This feel pretty apt.

Issue #841, opened 2019-10-15
Add configuration option to auto-camelCase requested flag names


thomassvensen:
We started using Launch Darkly before pulling in flopflip. In Launch Darkly, all our flags are defined as "kebab case", e.g. `this-is-a-flag`. Flopflip converts all of these to camelCase, e.g. `thisIsAFlag`, as you know. But we really want to keep flag names as kebab case in our code, as it much easier to mentally connect the code and LD when the casing is the same. It is also, for example, much easier to search for references inside the code when we can just copy/paste from LD, rather then have to convert casing.

Currently, we work around the above problem by wrapping *every* call to flipflop with a `camelCase('this-is-a-flag')`.

A very nice solution would be if flopflip could do this for us. It could be a configuration parameter, or it could just be done for any received flag name. As flopflip always normalizes to camelCase after receiving from LD, it actually seems quite safe to just do the same when receiving requests from the client as well? But a (opt-in) config parameter is of course a safer route to avoid breaking any existing code.

tdeekens:
Thanks for the issue. As flopflip supports multiple adapters the flag format can vary. As a consequence we decided to use camel casing initially. It felt more naturally to write e.g. when accessing flag values or keeping a flag name constant similar to its content. I agree thar in some caes, e.g. to lower the mental over head when looking at LD, kebab would be easier. An option would be to accept a casing argument for flopflip when configuring which continues to default to camel case to remain backwards compatible.

jstclair:
I think the original point wasn't the internal storage, but rather when you request a flag:

```jsx
<ToggleFeature flag={camelCase('our-flag-name')}>
```
or 
```jsx
injectFeatureToggles([
    camelCase('our-flag-a'),
    camelCase('our-flag-b'),
  ])
```
Just a quick look through the code definitions, it seems that perhaps adding a transform function to the AdapterArgs passed to `createFlopFlipEnhancer` (in this case, one could pass `{ transformer: camelCase }` but the default could be identity `( transformer: input => input }`. Naming is bad, since this would more accurately be `getFlagNameTransformer`

tdeekens:
That sounds reasonable yes but I'd like to not have this burdan on the consumer of flopflip to repeat the "mapping into internal flag represenations". 

I was wondering if something like:

```jsx
<ConfigureFlopFlip
  adapter={adapter}
  adapterArgs={{ clientSideId, user, flagCasing: 'camel' | 'kebab' }}
  render={() => <App />}
/>
[...]

PR #845, opened 2019-10-18
refactor: to automatically normalise flags

tdeekens:
#### Summary

This pull request takes on #841 to normalise flags by default. 

I still have to reflect on the consequences of this and how to make adapters backwards compatible.

E.g. if you have non normalised flags in local storage we need to normalise to forward their value correctly.

I like this change but it's a bit more risky than it seems ;)


tdeekens:
Ok, this should do the trick now. All flag state in all adapters is normalized and when you pass in some non-normalized flag name we also map into it.

PR #890, opened 2019-12-27
fix(types): adapter children is either a node or a function

emmenko:
#### Summary

The type declaration of `ConfigureAdapterChildren` is not entirely correct.

#### Description

The `ConfigureAdapter` component of the `react` package is typed with `children` being

```tsx
type ConfigureAdapterChildren = ({
  isAdapterReady,
}: {
  isAdapterReady: boolean;
}) => React.ReactNode | React.ReactNode | null;
```

which makes it to always be a function. As a consequence, passing a `ReactNode` as a child, results in a type error.
I think the mistake is in the syntax `=> React.ReactNode | React.ReactNode | null`, which hints that the **return type** of the function is either one of the listed types, instead of the `ConfigureAdapterChildren` type being etiher-or.

To fix that, I split the types so that the union is less ambiguous.

### DIFF
--- a/packages/react/modules/components/configure-adapter/configure-adapter.tsx
+++ b/packages/react/modules/components/configure-adapter/configure-adapter.tsx
@@ -8,6 +8,7 @@ import {
   AdapterReconfiguration,
   AdapterReconfigurationOptions,
   ConfigureAdapterChildren,
+  ConfigureAdapterChildrenFn,
 } from '@flopflip/types';
 import AdapterContext, { createAdapterContext } from '../adapter-context';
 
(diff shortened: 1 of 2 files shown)


packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -244,20 +249,20 @@ export default class ConfigureAdapter extends React.PureComponent<
...
-          const isAdapterReady: boolean = this.props.adapter.getIsReady();
+          const isAdapterReady = this.props.adapter.getIsReady();
 
           if (isAdapterReady) {
             if (typeof this.props.render === 'function')
               return this.props.render();
           }
 
-          if (typeof this.props.children === 'function')
+          if (isFunctionChildren(this.props.children))
```
emmenko: This is a consequence of the "correct" union type, as TS cannot infer the correct type

image

PR #892, opened 2019-12-27
fix(types): do not expose event handlers to adapterArgs props

emmenko:
#### Summary

According to the types of `adapterArgs` prop, the `ConfigureFlopFlip` component currently requires the user to pass props such as `onStatusStateChange` and `onFlagsStateChange`.

As far as I understand, those props are only used internally and therefore do no need to be exposed to the `ConfigureFlopFlip` component.

### DIFF
--- a/packages/react-broadcast/modules/components/configure/configure.tsx
+++ b/packages/react-broadcast/modules/components/configure/configure.tsx
@@ -5,6 +5,7 @@ import {
   Flags,
   Adapter,
   AdapterArgs,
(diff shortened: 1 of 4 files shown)


emmenko:
Btw, not sure why the "prerelease" action is running on my fork 🤔
https://github.com/emmenko/flopflip/runs/365268609

image

tdeekens:
@emmenko fixed now. Releasing will only run on my base.

PR #913, opened 2020-01-05
refactor: to remove hooks check

tdeekens:


packages/react-broadcast/modules/hooks/use-adapter-status/use-adapter-status.ts
```diff
@@ -5,16 +5,10 @@ import {
   AdapterStatus as AdapterStatusType,
 } from '@flopflip/types';
 
-export default function useAdapterStatus(): Error | AdapterStatusType {
-  if (typeof React.useContext === 'function') {
-    const adapterContext: AdapterContextType = React.useContext(AdapterContext);
+export default function useAdapterStatus(): AdapterStatusType {
```
tdeekens: In this case I value consitency. If this is inted ed to change we can do separately :)

PR #946, opened 2020-01-19
refactor(react): configure adapter to react.fc

tdeekens:


packages/react/modules/components/configure-adapter/configure-adapter.spec.js
```diff
@@ -15,21 +16,47 @@ const createTestProps = ({ adapter }) => ({
       key: 'foo-user-key',
     },
   },
+  adapterStatus: AdapterStates.CONFIGURED,
```
tdeekens: We can overwrite this under test.

packages/react/modules/components/configure-adapter/configure-adapter.spec.js
```diff
@@ -15,21 +16,47 @@ const createTestProps = ({ adapter }) => ({
...
   onFlagsStateChange: jest.fn(),
   onStatusStateChange: jest.fn(),
   adapter,
 });
 
-const TestComponent = () => <span>Test component</span>;
+const TestComponent = props => {
+  const adapterContext = React.useContext(AdapterContext);
```
tdeekens: We use this beefed up test component to test more things.

packages/react/modules/components/configure-adapter/configure-adapter.spec.js
```diff
@@ -15,21 +16,47 @@ const createTestProps = ({ adapter }) => ({
...
 const render = ({ props, adapter }) => {
   const baseProps = createTestProps({ adapter });
   const mergedProps = { ...baseProps, ...props };
 
   const rendered = rtlRender(<ConfigureAdapter {...mergedProps} />);
-  const waitUntilReady = () => Promise.resolve();
 
-  return { ...rendered, waitUntilReady, props: mergedProps };
+  const waitUntilStatus = (status = AdapterStates.CONFIGURED) =>
+    rendered.findByText(`Is ${status}: Yes`);
```
tdeekens: Maybe these should be waitUntilConfigured etc but it's easier this way. The downside is that rtl shows errors with non interpolate vairables.

packages/react/modules/components/configure-adapter/configure-adapter.spec.js
```diff
@@ -38,27 +65,27 @@ describe('rendering', () => {
       it('should invoke render prop', async () => {
         const adapter = createAdapter();
         adapter.getIsReady.mockReturnValue(true);
-        const props = { render: jest.fn() };
+        const props = { render: jest.fn(() => <TestComponent />) };
```
tdeekens: Otherwise we wait for nothing cause we render nothing :)

tdeekens:
Tests should pass now. Bloody hell. Getting a library to be ready for Concurrent Mode in React is more work than one initially expects. It affects how to test things quite a bit.

packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -42,91 +42,102 @@ type Props = {
...
-     */
-    this.setState(
-      {
-        appliedAdapterArgs: nextAdapterArgs,
-      },
-      () => {
-        this.pendingAdapterArgs = null;
-      }
-    );
+  const getIsAdapterConfigured = React.useCallback(
```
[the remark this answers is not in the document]
tdeekens: I am actually trying to move away from booleans in the code base. They seemsed apt but easily get hairy too. Cause you all of a sudden end up at 2-3 of them and then have to do the `if isA && isB || isC` stuff instead of enum style things which are a bit more like state machine states.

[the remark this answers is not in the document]
emmenko: Ah ok, so this is covered in the other PR?

tdeekens:
Thanks @emmenko made the worthwhile adjustments suggested 👍

packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -20,18 +20,14 @@ import AdapterContext, { createAdapterContext } from '../adapter-context';
 
 type valueof<T> = T[keyof T];
 
-type AdapterStates = {
-  UNCONFIGURED: string;
-  CONFIGURING: string;
-  CONFIGURED: string;
-};
-export const AdapterStates: AdapterStates = {
+export const AdapterStates = {
```
emmenko: Nit: variables should start with lowercase `adapterStates`

packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -20,18 +20,14 @@ import AdapterContext, { createAdapterContext } from '../adapter-context';
...
   UNCONFIGURED: 'unconfigured',
   CONFIGURING: 'configuring',
   CONFIGURED: 'configured',
-};
+} as const;
+export type TAdapterStates = typeof AdapterStates[keyof typeof AdapterStates];
```
emmenko:
```suggestion
export type TAdapterStates = ValueOf<typeof adapterStates>;
```

packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -42,91 +38,104 @@ type Props = {
   render?: () => React.ReactNode;
   children?: TConfigureAdapterChildren;
 };
-type State = {
-  appliedAdapterArgs: TAdapterArgs;
-};
-type AdapterState = valueof<AdapterStates>;
+type TAdapterState = valueof<TAdapterStates>;
```
emmenko: I think this is a duplicate

packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -20,18 +20,14 @@ import AdapterContext, { createAdapterContext } from '../adapter-context';
 
 type valueof<T> = T[keyof T];
```
emmenko: Type utilities should be written with capital letter: 

```suggestion
type ValueOf<T> = T[keyof T];
```

packages/react/modules/components/configure-adapter/configure-adapter.tsx
tdeekens: Yeah, this is me being weird however. It's strange to ready `adapterStates.current = adapterStates.PENDING`. I agree that only new-able things should be capitalized. I just subjectively like it like this :D

packages/react/modules/components/configure-adapter/configure-adapter.tsx
tdeekens: Oh superb catch!

packages/react/modules/components/configure-adapter/configure-adapter.tsx
emmenko: Ah I see. Yeah fine :)

Issue #942, opened 2020-01-18
Ability to pass attributes to SpiltIO adapter


jeffreyffs:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to pass custom attributes to the SplitIO Adapter so I can configure my feature flags based on those custom attributes (ie: user platform (ios/android/web)).

see: https://help.split.io/hc/en-us/articles/360020448791-JavaScript-SDK#attribute-syntax

**Describe the solution you'd like**
I believe we would need to update the SplitioAdapterArgs type, save the new adapter args into adapter state somewhere around [here](https://github.com/tdeekens/flopflip/blob/master/packages/splitio-adapter/modules/adapter/adapter.ts#L191)

and then pass it into `getTreatments` [here](https://github.com/tdeekens/flopflip/blob/master/packages/splitio-adapter/modules/adapter/adapter.ts#L89) and [here](https://github.com/tdeekens/flopflip/blob/master/packages/splitio-adapter/modules/adapter/adapter.ts#L141)

**Describe alternatives you've considered**
Not many alternatives, the attributes have to get into the `getTreatments` call some how :)

If the above makes sense, I can take a stab at making a PR

tdeekens:
Thanks for reporting. That sounds about right. The adapters differ a bit as to where the underlying SDK expects the user attributes. I actually do not use splitio myself. The adapter is a community maintained addition. Having said that: would you take a shot at a PR?

jeffreyffs:
Sounds good. I'll take a poke at it sometime this week :)

jeffreyffs:
Tested this out today and it works great. Thanks!

tdeekens:
Cool thanks for letting me know. Lots of other small improvements went out with the last release 🤖

PR #954, opened 2020-01-20
feat(splitio): support custom attributes

jeffreyffs:
#### Summary

Added the ability to pass custom attributes into splitio `getTreatments` function. Solves #942 

#### Description

Add a new object under `options` that is `treatmentAttributes`. It will key spread into `getTreatments` along side the `key` from user.

#### Technical debt & future

I don't love copying the type into `@flopflip/types` for SplitIO.Attributes.. but I think it is better than `any`. Open to suggestions.

On a side note, I couldn't figure out how to get all the tests to pass locally. Received a lot of errors like:
```
packages/react-redux/modules/hooks/use-flag-variations/use-flag-variations.spec.js
  ● Test suite failed to run

    Cannot find module '@flopflip/react' from 'use-flag-variations.ts'

    However, Jest was able to find:
        './use-flag-variations.spec.js'
        './use-flag-variations.ts'
```

but the splitio adapter test passed :)

### DIFF
--- a/packages/splitio-adapter/modules/adapter/adapter.ts
+++ b/packages/splitio-adapter/modules/adapter/adapter.ts
@@ -28,6 +28,7 @@ type AdapterState = {
     onStatusStateChange: OnStatusStateChangeCallback;
   };
   splitioSettings?: SplitIO.IBrowserSettings;
+  treatmentAttributes?: SplitIO.Attributes;
 };
 
 const adapterState: AdapterState = {
@@ -86,10 +87,10 @@ const subscribeToFlagsChanges = ({
   if (adapterState.client) {
     adapterState.client.on(adapterState.client.Event.SDK_UPDATE, () => {
       if (adapterState.client) {
-        const flags = adapterState.client.getTreatments(
-          flagNames,
-          adapterState.user as SplitIO.Attributes
-        );
+        const flags = adapterState.client.getTreatments(flagNames, {
+          ...adapterState.user,
+          ...adapterState.treatmentAttributes,
+        } as SplitIO.Attributes);
 
         onFlagsStateChange(normalizeFlags(flags));
       }
@@ -138,10 +139,10 @@ const subscribe = ({
 
         if (adapterState.client && adapterState.manager) {
           flagNames = adapterState.manager.names();
-          flags = adapterState.client.getTreatments(
-            flagNames,
-            adapterState.user as SplitIO.Attributes
-          );
+          flags = adapterState.client.getTreatments(flagNames, {
+            ...adapterState.user,
+            ...adapterState.treatmentAttributes,
+          } as SplitIO.Attributes);
 
           onFlagsStateChange(normalizeFlags(flags));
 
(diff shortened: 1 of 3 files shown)


jeffreyffs:
I decided the `treatmentAttributes` didn't feel great in the `options` key, so I moved it to the top level of AdapaterArgs for SplitIO.

tdeekens:
Sounds good yeah. I will have a look later this week. Maybe locally things fail as you haven't built all packages?

jeffreyffs:
I did a `lerna bootstrap`, but I couldn't find any documentation for how to get started developing for this package, so I'm not sure if I missed something.

tdeekens:
It's setup with yarn workspaces though. Lerna is just used for the publishing and building (the latter isn't needed anymore even since later workspace additions).

jeffreyffs:
ah, good to know :) I saw the `lerna` badge and thought it was managed by `lerna`. I ran `yarn build`, which looked like it went and built everything. Most of the tests are passing locally now, except for a few complaining about `Object.fromEntries`:

```
● injectFeatureToggles › with `propKey` › should have feature disabling prop for `disabledFeature`

    TypeError: Object.fromEntries is not a function

      20 |   const WrappedComponent = (ownProps: OwnProps) => {
      21 |     const flagVariations = useFlagVariations(flagNames);
    > 22 |     const flags = Object.fromEntries(
         |                          ^
      23 |       flagNames.map((flagName, indexOfFlagName) => [
      24 |         flagName,
      25 |         flagVariations[indexOfFlagName],
```

tdeekens:
I would assume maybe that your node version doesn’t support it. Just a guess...

jeffreyffs:
Ah, correct. I was using node 10 for another project. Thanks for the help! All tests passing locally now

packages/splitio-adapter/modules/adapter/adapter.ts
```diff
@@ -186,9 +187,16 @@ class SplitioAdapter implements SplitioAdapterInterface {
...
-    const { authorizationKey, user, options = {} } = adapterArgs;
+    const {
+      authorizationKey,
+      user,
+      options = {},
+      treatmentAttributes,
+    } = adapterArgs;
 
     adapterState.user = ensureUser(user);
+    adapterState.treatmentAttributes = treatmentAttributes;
```
tdeekens: Should these treatments also be changable from `reconfigure`?

jeffreyffs: Hm, good point. I'll get an update up probably tomorrow. Thanks for the quick review!

jeffreyffs: done!

packages/splitio-adapter/modules/adapter/adapter.ts
```diff
@@ -220,9 +228,23 @@ class SplitioAdapter implements SplitioAdapterInterface {
       );
     }
 
+    let hasDifference = false;
```
tdeekens: Nit: It feels easier to not have a `let` here but instead 

```js
const hasUserChanged = !isEqual(adapterState.user, adapterArgs.user);
const hasTreatmentChanged = !isEqual(
  adapterState.treatmentAttributes,
  adapterArgs.treatmentAttributes
);

if (hasUserChanged) { adapterState.user = ensureUser(adapterArgs.user); }
if (hasTreatmentChanged) { adapterState.treatmentAttributes = adapterArgs.treatmentAttributes; }

if (hasUserChanged && hasTreatmentChanged && adapterState.manager && adapterState.client) {
   adapterState.client.destroy();
}

jeffreyffs: hmm, that probably needs to be `(hasUserChanged || hasTreatmentChanged && (adapterState.manager && adapterState.client))` as we need to destroy the client if either of those change. But I agree, this looks nicer.

tdeekens: True. Sorry.

jeffreyffs: No need to apologize. Appreciate the feedback :)

Updated.

tdeekens:
Thanks for your contribution. It will be in the next release :)

tdeekens:
Published a canary for you to try it out.

Successfully published:
 - @flopflip/splitio-adapter@1.6.1-next.55+4236147

Issue #829, opened 2019-09-30
LaunchDarkly Client passthrough


timothyjoh:
This is related to the issue #713 but the request is to expose the ldClient so that I CAN issue the tracking events manually.

I can't always rely on the `fetchGoals` getting me the tracking data I want, so we need to expose the `LDClient` (which is imported in the adapter on line 17) through to the `ToggleFeature` component. 

It would be good if when using the "Function as a Child" approach, it was passed in along with the `isFeatureEnabled` boolean, and in the case of child Components, passed down as a prop to each of the Toggled & UnToggled components.

tdeekens:
Thanks for the issue. Seems reasonable to pass it down for the LD adapter. What would you think of a useAdapter hook? It would allow me not to touch other components.

timothyjoh:
After writing this, I ended up forking the code, to make a change to the adapter, exposing the client

https://github.com/timothyjoh/flopflip/commit/5ad7acbdbb9733fedd64dd32e59025544c5e487e

that way, all I had to do was write an
```
adapter.getClient() &&
        adapter.getClient().track("Custom Event Here");
```
in a callback inside my react component. 

Do you think this is a cleaner way of exposing it? I certainly think this is a) less code and b) cleaner approach.

tdeekens:
Thanks for getting back. Yes this would be needed regardless and is what I would have added too. Additionally one could write a nice hook or other React based integration so application level code just runs useAdapterClient which hides the detail of what client it is and where it comes fro,

timothyjoh:
I guess the problem here is that each vendors different Client may have different method signatures on it. For LD, I need `track(str)` but for split.io or Optimizely (we are considering using them) it would/may be different.

If we do work on Optimizely, I will be sure to commit an adapter back for that ;)

tdeekens:
Yes, that is true but at least you're not repeating the adapter you pass into LD and get out.

tdeekens:
I think I could build the API but I still would like you opinion on the following. Exposing the client behind an adapter changes the API contract of this library quite drastically. Right now we can update the clients (e.g. LD) without you ever as a consumer ever having to worry. As we encapsulate the client's API for you. Any breaking changes in the clients' APIs will be taken care of by this library - more specifically the respective adapter. By exposing the underlying client however we would have to mark any breaking change of the client as a breaking change of the adapter as we don't know how you might use the client (or what API on it). Does this make sense? WDYT?

tdeekens:
[...]

PR #969, opened 2020-02-03
feat(launch-darkly): expose launchdarkly client

zslabs:
#### Summary

Exposes LaunchDarkly client

#### Description

https://github.com/tdeekens/flopflip/issues/829

This allows us to call LD's `track()` method (as an example).

cc tdeekens

#### Technical debt & future

1. Keeping assumption that FlopFlip will control adapter version for now.

### DIFF
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.ts
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.ts
@@ -231,6 +231,10 @@ class LaunchDarklyAdapter implements LaunchDarklyAdapterInterface {
     return adapterState.isReady;
   }
 
+  getClient(): LDClient | undefined {
+    return adapterState.client;
+  }
+
   getFlag(flagName: FlagName): FlagVariation | undefined {
     return adapterState.flags[flagName];
   }
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
@@ -41,6 +41,10 @@ describe('when configuring', () => {
     expect(adapter.getIsReady()).toBe(false);
   });
 
+  it('should not return client', () => {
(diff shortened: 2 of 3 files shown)


packages/launchdarkly-adapter/modules/adapter/adapter.ts
```diff
@@ -231,6 +231,10 @@ class LaunchDarklyAdapter implements LaunchDarklyAdapterInterface {
     return adapterState.isReady;
   }
 
+  getClient(): LDClient | undefined {
```
tdeekens: Would you prefer the client to be exposed fully or just the track function?

zslabs: Well - I figured my more isolated usecase of just needing the track function may not be what others may need in the future - so felt returning the entire client would save us the headache of needing to do a follow-up PR.

tdeekens: Yup. However, at the cost that in theory and breaking change to the client's release cadence, we don't control, being a breaking change to this package. But I don't mind syncing those by now.,

zslabs: Thanks for getting this in! Really looking forward to having this released. Is that something I can submit a PR for as well by chance?

tdeekens: Let me see if canary releases still with with GitHub Actions. I saw some misconfig from my end lately. Otherwise I can publish it later.

tdeekens:
Published a canary for you to try it out:

Successfully published:
 - @flopflip/launchdarkly-adapter@2.10.1-next.55+4236147

zslabs:
This is indeed working! I now have access to the track method and was able to verify my clicks are getting back to Launch Darkly via their debugger and inside of a feature flag's "Experiments" tab.

PR #977, opened 2020-02-11
refactor: let TS infer return types if possible, fix props for ToggleFeature

emmenko:
#### Summary

Most of the return types are usually inferred by TypeScript, there is no need to explicitly define them. This is also important to get the "correct" return types for React components.

Additionally, I fixed up a bit the `ToggleFeature` component props.

See inline comments for more detailed info.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -4,7 +4,6 @@ parserOptions:
   ecmaFeatures:
     jsx: true
     modules: true
-  project: './tsconfig.json'
 extends:
   - xo
   - xo-typescript
--- a/packages/launchdarkly-adapter/index.ts
+++ b/packages/launchdarkly-adapter/index.ts
@@ -1,0 +1,5 @@
+// This file exists  because we want jest to use our non-compiled code to run tests
+// if this file is missing, and you have a `module` or `main` that points to a non-existing file
+// (ie, a bundle that hasn't been built yet) then jest will fail if the bundle is not yet built.
+// all apps should export all their named exports from their root index.js
+export * from './modules';
--- a/packages/react-redux/modules/components/toggle-feature/toggle-feature.tsx
+++ b/packages/react-redux/modules/components/toggle-feature/toggle-feature.tsx
@@ -1,20 +1,19 @@
 import { TFlagName, TFlagVariation } from '@flopflip/types';
 
 import React from 'react';
-import { ToggleFeature as SharedToggleFeature } from '@flopflip/react';
+import {
+  ToggleFeature as SharedToggleFeature,
+  TToggleFeatureProps,
+} from '@flopflip/react';
 import { useFeatureToggle } from '../../hooks/';
 
 type Props = {
   flag: TFlagName;
   variation: TFlagVariation;
-};
+} & Exclude<TToggleFeatureProps, 'isFeatureEnabled'>;
 
-const ToggleFeature = <OwnProps extends Props>(
-  props: OwnProps
-): React.ReactNode => {
+const ToggleFeature = <OwnProps extends Props>(props: OwnProps) => {
   const isFeatureEnabled = useFeatureToggle(props.flag, props.variation);
-
-  // @ts-ignore
   return <SharedToggleFeature {...props} isFeatureEnabled={isFeatureEnabled} />;
 };
 
--- a/packages/react/modules/components/toggle-feature/toggle-feature.ts
+++ b/packages/react/modules/components/toggle-feature/toggle-feature.ts
@@ -2,20 +2,19 @@ import React from 'react';
 import warning from 'tiny-warning';
 import { isValidElementType } from 'react-is';
 
-type Props = {
+type RenderFnArgs = {
+  isFeatureEnabled: boolean;
+};
+type RenderFn = (args: RenderFnArgs) => React.ReactNode;
+export type Props = {
   untoggledComponent?: React.ComponentType;
   toggledComponent?: React.ComponentType;
   render?: () => React.ReactNode;
-  children?:
-    | (({ isFeatureEnabled: boolean }) => React.ReactNode)
-    | React.ReactNode;
+  children?: RenderFn | React.ReactNode;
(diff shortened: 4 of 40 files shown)


.eslintrc.yaml
```diff
@@ -4,7 +4,6 @@ parserOptions:
   ecmaFeatures:
     jsx: true
     modules: true
-  project: './tsconfig.json'
```
emmenko: As far as I know, this is the default, there is no need to define it.
Of course I can revert it if you want to keep it.

packages/launchdarkly-adapter/index.ts
```diff
@@ -0,0 +1,5 @@
+// This file exists  because we want jest to use our non-compiled code to run tests
+// if this file is missing, and you have a `module` or `main` that points to a non-existing file
+// (ie, a bundle that hasn't been built yet) then jest will fail if the bundle is not yet built.
+// all apps should export all their named exports from their root index.js
+export * from './modules';
```
emmenko: As we do in appkit, this "proxy" file is necessary to allow typescript to find the exports of a package, if no `dist` is available.

packages/react-redux/modules/components/toggle-feature/toggle-feature.tsx
```diff
@@ -1,20 +1,19 @@
 import { TFlagName, TFlagVariation } from '@flopflip/types';
 
 import React from 'react';
-import { ToggleFeature as SharedToggleFeature } from '@flopflip/react';
+import {
+  ToggleFeature as SharedToggleFeature,
+  TToggleFeatureProps,
```
emmenko: The underlying props of the `ToggleFeature` component are now exported, so that we can "merge" them with the props from the "higher level" `ToggleFeature` components.
This should fix the issue with the "missing children", etc"

packages/react/modules/components/toggle-feature/toggle-feature.ts
```diff
@@ -2,20 +2,19 @@ import React from 'react';
...
+};
+type RenderFn = (args: RenderFnArgs) => React.ReactNode;
+export type Props = {
   untoggledComponent?: React.ComponentType;
   toggledComponent?: React.ComponentType;
   render?: () => React.ReactNode;
-  children?:
-    | (({ isFeatureEnabled: boolean }) => React.ReactNode)
-    | React.ReactNode;
+  children?: RenderFn | React.ReactNode;
```
emmenko: Just extracted the fn into a type

packages/react/modules/components/toggle-feature/toggle-feature.ts
```diff
@@ -32,8 +31,13 @@ const ToggleFeature = (props: Props): React.ReactNode => {
     if (props.toggledComponent)
       return React.createElement(props.toggledComponent);
 
-    if (props.children && !isEmptyChildren(props.children))
-      return React.Children.only(props.children);
+    if (props.children) {
+      if (typeof props.children === 'function')
+        return props.children({
```
emmenko: Maybe I got this wrong, but I think we should render here the children as a function, in case it's a function.

packages/react/modules/components/toggle-feature/toggle-feature.ts
```diff
@@ -32,8 +31,13 @@ const ToggleFeature = (props: Props): React.ReactNode => {
...
+          isFeatureEnabled: props.isFeatureEnabled,
+        });
+      return React.Children.only<React.ReactNode>(props.children);
```
emmenko: I'm not sure what the `isEmptyChildren` check is for, so I left it out as we want to render the children here.

Please correct me if I got this wrong.

packages/react/modules/components/toggle-feature/toggle-feature.ts
```diff
@@ -2,20 +2,19 @@ import React from 'react';
...
   toggledComponent?: React.ComponentType;
   render?: () => React.ReactNode;
-  children?:
-    | (({ isFeatureEnabled: boolean }) => React.ReactNode)
-    | React.ReactNode;
+  children?: RenderFn | React.ReactNode;
   isFeatureEnabled: boolean;
 };
 
-const isEmptyChildren = (children: React.ReactNode): boolean =>
```
emmenko: By looking at the code I felt that this is not necessary (see comments below)

packages/react/modules/hocs/wrap-display-name/wrap-display-name.ts
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
 
-export default (
+export default function wrapDisplayName(
```
emmenko: The name makes it "easier" to "inspect" the type signature "on hover"

tsconfig.json
```diff
@@ -27,6 +27,7 @@
     "allowJs": false
   },
   "include": [
+    "packages/**/*.ts",
```
emmenko: This is to include the "proxy exports" in each package

packages/react/modules/components/toggle-feature/toggle-feature.ts
tdeekens: This happens in L45 if I am not mistaken?

packages/react/modules/components/toggle-feature/toggle-feature.ts
tdeekens: Can't fully recall. Was something people recommended doing. I am fine with leaving it out!

packages/react/modules/components/toggle-feature/toggle-feature.ts
emmenko: In both yes. But we need to duplicate it as otherwise we might not get to L45 (for example if the render prop is also defined).
Unless you don't bother about this case.

packages/react/modules/components/toggle-feature/toggle-feature.ts
tdeekens: True.

tdeekens:
Thanks again!

PR #980, opened 2020-02-12
feat: add proper handling of flag subscription instead of component mount status

tdeekens:
#### Summary

This pull request changes the logic of how state propagates by having adapters being able to subscribe and unsubcribe over keeping track of mount status.


packages/react/modules/hooks/use-adapter-subscription/use-adapter-subscription.ts
```diff
@@ -0,0 +1,18 @@
...
+    if (adapter.subscribe) {
+      adapter.subscribe();
+    }
+
+    return () => {
+      if (adapter.unsubscribe) {
+        adapter.unsubscribe();
+      }
+    };
+  }, [adapter]);
```
emmenko: Maybe better to destructure the subscribe/unsubscribe methods and use those in the dependency list.

tdeekens: You mean like 9124ee9, right? It won't help much I think as the adapter itself doesn't change. It's a static import. Unlike props which change as a whole when one prop changes.

I am really unsure what the cause for problems here is. Does the implementation roughly look fine to you? The errors come when we e.g. test `yarn test:watch use-feature-toggle`. Only the broadcast package has it as redux takes care of the it otherwise (the setters are action creators).

packages/react/modules/hooks/use-adapter-subscription/use-adapter-subscription.spec.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { render as rtlRender } from '@flopflip/test-utils';
+import useAdapterSubscription from './use-adapter-subscription';
+
+const createAdapter = () => ({
+  getIsReady: jest.fn(() => false),
+  configure: jest.fn(() => Promise.resolve()),
+  reconfigure: jest.fn(() => Promise.resolve()),
+  subscribe: jest.fn(() => Promise.resolve()),
```
emmenko: Those are not promises, right?

tdeekens: My bad.

tdeekens:
> Thanks, makes sense to me!

To me too. It just doesn’t work. There are still logs. It is just that I don’t throw here (yet).

emmenko:
What does not work?

tdeekens:
React still reports that it receives state updates after components have been unmounted. Also visible on [CI](https://github.com/tdeekens/flopflip/pull/980/checks?check_run_id=441133576). 

CleanShot 2020-02-12 at 17 58 38@2x

I tried debugging it a bit but ended up nowhere so far.

tdeekens:
However with something like this being logged

CleanShot 2020-02-12 at 18 36 19@2x

That there seem to be competing set/get calles. So we set to true, then set to false and then get the false while be would like true. Maybe that is cause multiple adapters are being rendered at the same time and as a result one adapter can not maintain subscription state for multiple configure react components.

emmenko:
Which test is logging the error?

> Maybe that is cause multiple adapters are being rendered at the same time

Can you elaborate? Is it only in the tests or is it a "normal" thing?

tdeekens:
Most tests are. For instance those of use-feature-toggle in react-broadcast. It is visible on CI. 

It should only happen on CI. However, also if we use it in our work code base we run into it. The adapters are singletons. Their state is too. So if multiple configure components are rendered they compete in a state-battle with the singleton adapter messing each others state up. E.g configure-1 unsubscribes, then configure-2 subscribes again. Now a setState is triggered through the callbacks: both configure components receive it even if 1 unsubscribed as it unmounted. 

That is my current reasoning. There are solutions to this. Just not sure yet what a nice one is 🦊. 

Does this make sense?

emmenko:
Hmm ok so in production usage there should only be one "configure component", so everything works fine. However, in tests scenarios there might be multiple ones and therefore we see race conditions.

Did I get it right?

Maybe a follow up question: do the adapters need to be singletones? What was the reason for this?

tdeekens:
Yes, that‘s how I see it top know. As to why? Simplicity. Also splitio or LD adapters would usually not exist as multiple instances in one app. Also the local storage adapter operates on the same state even if two exist. So there was never a real need until... 🙃. Maybe only the memory adapter can be cloned or something. As it is the one being used in test. So that on every rtl render you just clone/create new one from the existing one. While always one exists so it is not a breaking change 😒😩

packages/launchdarkly-adapter/modules/adapter/adapter.ts
```diff
@@ -31,7 +32,7 @@ type LaunchDarklyAdapterState = {
 const adapterState: TAdapterStatus & LaunchDarklyAdapterState = {
   isReady: false,
   isConfigured: false,
-  isUnsubscribed: false,
+  subscriptionStatus: TAdapterSubscriptionStatus.Subscribed,
```
tdeekens: Screw booleans. Another nite the others will become enums too. Sadly they can't be const enums due to babel which affects bundle size a bit.

packages/react-broadcast/modules/components/configure/configure.tsx
```diff
@@ -22,36 +25,56 @@ type State = {
   configurationId?: string;
 };
 
+const initialAdapterStatus: TAdapterStatus = {
+  isReady: false,
+  subscriptionStatus: TAdapterSubscriptionStatus.Subscribed,
+};
+const initialFlags: TFlags = {};
```
tdeekens: TypeScript rightfully, but all of a sudden, got me here.

packages/react-broadcast/modules/components/configure/configure.tsx
```diff
@@ -22,36 +25,56 @@ type State = {
...
         ...flags,
       }));
     },
-    [setFlags]
+    [setFlags, getHasAdapterSubscriptionStatus]
   );
 
   const handleUpdateStatus = React.useCallback(
-    (status: TAdapterStatus) => {
+    (status: TAdapterStatusChange) => {
```
tdeekens: This is more accurate as a type cause it's a partial.

packages/react/modules/hooks/use-adapter-subscription/use-adapter-subscription.ts
```diff
@@ -1,20 +1,34 @@
-import { TAdapter } from '@flopflip/types';
+import { TAdapter, TAdapterSubscriptionStatus } from '@flopflip/types';
 import React from 'react';
 
 function useAdapterSubscription(adapter: TAdapter) {
+  const useAdapterSubscriptionStatusRef = React.useRef(
+    TAdapterSubscriptionStatus.Subscribed
+  );
```
tdeekens: This is the trick. For know we duplicate this state in the react component. Making it easier for consumers cause nothing changes.

packages/react-broadcast/modules/components/configure/configure.tsx
emmenko: Nit, to make it "consistent" with the `useState` type

```
const initialFlags: State['flags'] = {};
```

packages/react-broadcast/modules/components/configure/configure.tsx
```diff
@@ -22,36 +25,56 @@ type State = {
   configurationId?: string;
 };
 
+const initialAdapterStatus: TAdapterStatus = {
```
emmenko: Nit, to make it "consistent" with the useState type

```
const initialAdapterStatus: State['status'] = {
```

packages/react/modules/hooks/use-adapter-subscription/use-adapter-subscription.ts
```diff
@@ -1,20 +1,34 @@
-import { TAdapter } from '@flopflip/types';
+import { TAdapter, TAdapterSubscriptionStatus } from '@flopflip/types';
 import React from 'react';
 
 function useAdapterSubscription(adapter: TAdapter) {
```
emmenko: Nit: `useGetAdapterSubscriptionStatus`?

packages/react/modules/hooks/use-adapter-subscription/use-adapter-subscription.ts
emmenko: Ah nice solution!

packages/react-broadcast/modules/components/configure/configure.tsx
```diff
@@ -22,36 +25,56 @@ type State = {
...
+  );
 
   // NOTE:
   //   Using this prevents the callbacks being invoked
   //   which would trigger a setState as a result on an unmounted
   //   component.
-  useAdapterSubscription(props.adapter);
+  const getHasAdapterSubscriptionStatus = useAdapterSubscription(props.adapter);
 
   const handleUpdateFlags = React.useCallback(
```
emmenko: Nit: you can specify the callback signature to `useCallback` instead of typing the function itself

```ts
React.useCallback<(foo: Foo) => void>(
  foo => {
    // foo is typed as Foo
  },
  []
)
```

packages/react/modules/hooks/use-adapter-subscription/use-adapter-subscription.ts
tdeekens: Yeah, wasn't sure myself. You don't have to use the getter and still something happens. The setup is more important (in Redux it's automatic). So I felt a more generic name is fine. But thanks for suggesting :)

tdeekens:
Thanks for the suggestions. Incorporated 3/4 155e8d6.

packages/react-broadcast/modules/components/configure/configure.tsx
tdeekens: This I am actually never super sure about. I've read articles where people say type the function as TS is smart enough to "fill in the generic" and the other way round. I haven't decides which way I like more yet. You have opinions on it? I kind of also see value in having generics about types allowing you to have it not in arguments directly.

emmenko: Yeah sometimes the function is "simple enough" that TS can infer the signature. I usually check if that's the case, otherwise I'll add the type annotations.
I kind of prefer to pass the signature to `useCallback` as it makes it easier to see the signature at first glance, especially if the function is big.

PR #981, opened 2020-02-13
fix: to build types

tdeekens:


tdeekens:
Worth a shot. Can't get worse I suppose :)

PR #991, opened 2020-02-17
fix: display name wrapping

tdeekens:


tdeekens:
I am too lazy to write a test and it's my library 😄

PR #995, opened 2020-02-23
refactor: to not use booleans for adapter state

tdeekens:
#### Summary

This pull request is a refactoring to _not_ use booleans as adapter status state.

#### Description

Booleans are hard to maintain over type. Better are enums. In that this PR:

1. Adds a `adapterConfigurationStatus` field on adapters and components
2. Adds a `TAdapterConfigurationStatus` to on the type level
3. Adds a backwards compatible layer to not have this being a breaking change
    - The `isReady` flag is mapped into in the selectors in all packages
4. Adapters get a `getIsAdapterStatus`-fn while mapping `getIsReady` with a warning onto it
5. All state in Redux and the Context (broadcast) is not cleaner while selectors map out of it

### DIFF
--- a/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
+++ b/packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
@@ -1,3 +1,4 @@
+import { TAdapterConfigurationStatus } from '@flopflip/types';
 import ldClient from 'launchdarkly-js-client-sdk';
 import adapter, { normalizeFlags } from './adapter';
 
@@ -20,11 +21,6 @@ const createClient = jest.fn(apiOverwrites => ({
 
   ...apiOverwrites,
 }));
-const AdapterStatus = {
-  Unconfigured: 0,
-  Configuring: 1,
-  Configured: 2,
-};
 
 const triggerFlagValueChange = (client, flagValue = false) =>
   client.on.mock.calls.forEach(([event, cb]) => {
@@ -43,9 +39,9 @@ describe('when configuring', () => {
   });
 
   it('should indicate that the adapter is not configured', () => {
-    expect(adapter.getIsConfigurationStatus(AdapterStatus.Configured)).toBe(
-      false
-    );
+    expect(
+      adapter.getIsConfigurationStatus(TAdapterConfigurationStatus.Configured)
+    ).toBe(false);
   });
 
   it('should not return client', () => {
@@ -173,7 +169,9 @@ describe('when configuring', () => {
       describe('when determining if adapter is ready', () => {
         it('should indicate that the adapter is configured', () => {
           expect(
-            adapter.getIsConfigurationStatus(AdapterStatus.Configured)
+            adapter.getIsConfigurationStatus(
+              TAdapterConfigurationStatus.Configured
+            )
           ).toBe(true);
         });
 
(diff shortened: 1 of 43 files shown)


tdeekens:
@emmenko I don't know who to ask otherwise but I'd need at least an 👁 or judgment on this 😄.

packages/launchdarkly-adapter/modules/adapter/adapter.spec.js
```diff
@@ -1,3 +1,4 @@
+import { TAdapterConfigurationStatus } from '@flopflip/types';
```
emmenko: Nit: this is supposed to be an enum, it's a bit weird to name is as a type. I'm not a naming expert but maybe we can make it less ambiguous.

For example: `AdapterConfigurationStatusEnum`

packages/types/index.ts
```diff
@@ -11,9 +11,13 @@ export enum TAdapterSubscriptionStatus {
   Subscribed,
   Unsubscribed,
 }
+export enum TAdapterConfigurationStatus {
+  Unconfigured,
+  Configuring,
+  Configured,
+}
```
emmenko: Nitpick: just a suggestion how I would name those values, based on my understanding so far.

```
NotConfigured
Pending
Configured
```

packages/localstorage-adapter/modules/adapter/adapter.ts
```diff
@@ -139,37 +140,42 @@ class LocalStorageAdapter implements TLocalStorageAdapterInterface {
...
       adapterState.emitter.emit(
         'flagsStateChange',
         normalizeFlags(storage.get('flags'))
       );
       adapterState.emitter.emit('statusStateChange', {
-        isReady: adapterState.isReady,
+        configurationStatus: adapterState.configurationStatus,
       });
-      adapterState.emitter.emit('readyStateChange');
+      adapterState.emitter.emit('__internalConfiguredStatusChange__');
```
emmenko: Nit: define `__internalConfiguredStatusChange__` as a constant variable somewhere?

packages/react/modules/components/adapter-context/adapter-context.ts
```diff
@@ -23,5 +23,22 @@ const createAdapterContext = (
...
+});
+
+const useAdapterContext = () => React.useContext(AdapterContext);
+
 export default AdapterContext;
-export { createAdapterContext };
+export {
+  createAdapterContext,
+  useAdapterContext,
+  selectAdapterConfigurationStatus,
```
emmenko: Can we combine the context and the selector?

```ts
const { isConfigured } = useAdapterConfigurationStatus(configurationStatus)
```

The `useAdapterContext` is then used internally in the `useAdapterConfigurationStatus` hook

packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -250,16 +251,20 @@ export default class ConfigureAdapter extends React.PureComponent<
...
+          if (isAdapterConfigured) {
             if (typeof this.props.render === 'function')
               return this.props.render();
           }
 
           if (isFunctionChildren(this.props.children))
             return this.props.children({
-              isAdapterReady,
+              // NOTE: Deprecated, please use `isAdapterConfigured`.
+              isAdapterReady: isAdapterConfigured,
```
emmenko: 💯

packages/react/modules/components/adapter-context/adapter-context.ts
tdeekens: Yeah thought about stuff like that but it got too late (brain fried). I also had to allow just passing in the selector `useAdapterContext(selectConfigurationStatus)` as in Redux.

tdeekens: The ideal goal would be to have the selectors for Redux and Broadcast the same. Not even shared maybe but similar to ease things.

PR #1000, opened 2020-02-25
refactor: to use debug value more

tdeekens:
#### Summary

Last nite fun.

Using `useDebugValue` can help with debugging and in the DevTools quite a bit. As a result the code also get's some leaner and more consise.

### DIFF
--- a/packages/react/modules/components/configure-adapter/configure-adapter.tsx
+++ b/packages/react/modules/components/configure-adapter/configure-adapter.tsx
@@ -40,12 +40,51 @@ type TProps = {
   children?: TConfigureAdapterChildren;
 };
 
-const ConfigureAdapter = (props: TProps) => {
+const useApplieAdapterArgsState = ({
+  initialAdapterArgs,
+}: {
+  initialAdapterArgs: TAdapterArgs;
+}) => {
   const [appliedAdapterArgs, setAppliedAdapterArgs] = React.useState<
     TAdapterArgs
-  >(props.adapterArgs);
-  const pendingAdapterArgs = React.useRef<TAdapterArgs | null>(null);
-  const adapterState = React.useRef<TAdapterStates>(AdapterStates.UNCONFIGURED);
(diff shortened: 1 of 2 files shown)


packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -64,7 +103,7 @@ const ConfigureAdapter = (props: TProps) => {
        */
       setAppliedAdapterArgs(nextAdapterArgs);
```
tdeekens: CleanShot 2020-02-25 at 09 44 32@2x
CleanShot 2020-02-25 at 09 44 24@2x

These now blow to be not callable. I don't follow why. Am I missing something? I would expect TS to infer the return type correctly. /cc @emmenko

packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -40,12 +40,51 @@ type TProps = {
   children?: TConfigureAdapterChildren;
 };
 
-const ConfigureAdapter = (props: TProps) => {
+const useApplieAdapterArgsState = ({
```
emmenko:
```suggestion
const useAppliedAdapterArgsState = ({
```

packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -64,7 +103,7 @@ const ConfigureAdapter = (props: TProps) => {
        */
       setAppliedAdapterArgs(nextAdapterArgs);
     },
-    []
```
emmenko: Any reason why this needed to change?

packages/react/modules/components/configure-adapter/configure-adapter.tsx
emmenko: Can't say much like this, I would need to check out the branch

PR #1002, opened 2020-02-26
fix: to use isEqual for localstorage adapter

tdeekens:


packages/localstorage-adapter/modules/adapter/adapter.ts
```diff
@@ -114,22 +115,25 @@ export const updateFlags = (flags: TFlags) => {
   adapterState.emitter.emit('flagsStateChange', nextFlags);
 };
 
+const didFlagsChange = (nextFlags: TFlags) => {
+  const previousFlags = adapterState.flags;
+
+  if (previousFlags === undefined) return true;
+
+  return isEqual(nextFlags, previousFlags);
```
ngbrown: should be a `!isEqual`, because the `true` from this function is true if the two flags are _not_ equal.

packages/localstorage-adapter/modules/adapter/adapter.ts
```diff
@@ -114,22 +115,25 @@ export const updateFlags = (flags: TFlags) => {
   adapterState.emitter.emit('flagsStateChange', nextFlags);
 };
 
+const didFlagsChange = (nextFlags: TFlags) => {
+  const previousFlags = adapterState.flags;
```
ngbrown: `adapterState.flags` is never set, so the object is always empty.

It looks like it should be set in `updateFlags`, `subscribeToFlagsChanges`, and `configure` right before each instance of `adapterState.emitter.emit('flagsStateChange', ... )`.

PR #1059, opened 2020-04-18
refactor: to only use import type

tdeekens:
#### Summary

This pull request refactors flopflip to only used `import type` where possible. A feature available since TypeScript 3.8.


emmenko:
😄🙌

PR #1061, opened 2020-04-18
refactor: to use only readonly types

tdeekens:
This pull request moves on with the new ESLint recommendation to only use readonly types. While doing so previously untyped parts of the code base are now also more strictly typed.

### DIFF
--- a/packages/react/modules/components/configure-adapter/configure-adapter.tsx
+++ b/packages/react/modules/components/configure-adapter/configure-adapter.tsx
@@ -1,3 +1,4 @@
+import type { DeepReadonly } from 'ts-essentials';
 import type {
   TFlags,
   TAdapter,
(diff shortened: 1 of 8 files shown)


packages/react/modules/components/configure-adapter/configure-adapter.tsx
```diff
@@ -181,6 +196,19 @@ const useHandleDefaultFlagsCallback = ({ onFlagsStateChange }) => {
...
 
+type TUseConfigurationEffectOptions = DeepReadonly<{
+  adapter: TAdapter;
+  shouldDeferAdapterConfiguration: TProps['shouldDeferAdapterConfiguration'];
+  getDoesAdapterNeedInitialConfiguration: TUseAdapterStateRefReturn['3'];
+  setAdapterState: TUseAdapterStateRefReturn['1'];
+  onFlagsStateChange: TAdapterEventHandlers['onFlagsStateChange'];
+  onStatusStateChange: TAdapterEventHandlers['onStatusStateChange'];
+  applyAdapterArgs: TUseAppliedAdapterArgsStateReturn['1'];
+  getAdapterArgsForConfiguration: TUsePendingAdapterArgsRefReturn['2'];
```
tdeekens, line 207: Dislike it. Should use `ReturnType`. Might do as a follow-up.

PR #1088, opened 2020-05-10
feat: setup changeset

tdeekens:


.github/workflows/release.yml
```diff
@@ -1,58 +1,72 @@
...
         run: yarn install --frozen-lockfile
-      - name: Publish to next dist-tag
-        if: success()
+
+      - name: Creating .npmrc
         run: |
-          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
-          yarn release:next
+          cat << EOF > "$HOME/.npmrc"
+            email=npmjs@commercetools.com
```
emmenko, line 54: 🙂

.github/workflows/release.yml
```diff
@@ -1,58 +1,72 @@
...
+            //registry.npmjs.org/:_authToken=$NPM_TOKEN
+          EOF
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        # uses: changesets/action@master
+        uses: emmenko/action@nm-fix-npmrc-overwrite
```
emmenko, line 63: PS: my PR got merged, you can use `changesets/action@master`

tdeekens: Yeah. Thanks. We need to use that on app-kit too. The branch was deleted. Breaks stuff :).  Pushed to master.

.github/workflows/release.yml
tdeekens: Danke.

.github/workflows/release.yml
emmenko: Ah right :)

PR #1235, opened 2020-09-06
chore: update deps

tdeekens:


packages/types/index.ts
```diff
@@ -1,9 +1,10 @@
+import type React from 'react';
 import type { LDClient as TLDClient } from 'launchdarkly-js-client-sdk';
 import type { DeepReadonly } from 'ts-essentials';
 
 export type TFlagName = string;
 export type TFlagVariation = boolean | string;
-export type TFlag = [TFlagName, TFlagVariation];
+export type TFlag = [flagName: TFlagName, flagVariation: TFlagVariation];
```
emmenko, line 7: FYI: this is a bit of a breaking change as it only works for TS v4

PR #1313, opened 2020-11-17
Fix the syntax on isAdapterConfigured

AlexHayton:
#### Summary

Fixed the syntax on one of the examples for isAdapterConfigured

#### Description

If you copy and paste from that example it doesn't work!


tdeekens (approved):
Oh thanks!

PR #1318, opened 2020-11-21
chore: update babel to test auto changeset addition

tdeekens:


.github/.renovaterc.json
```diff
@@ -16,6 +16,18 @@
       "updateTypes": ["minor", "patch"],
       "groupName": "all dependencies",
       "groupSlug": "all"
+    },
+    {
+      "updateTypes": ["major"],
+      "labels": ["changeset:patch"]
```
emmenko, line 22: `changeset:major`

.github/.renovaterc.json
```diff
@@ -16,6 +16,18 @@
...
+    },
+    {
+      "updateTypes": ["minor"],
+      "labels": ["changeset:patch"]
```
emmenko, line 26: `changeset:minor`

tdeekens: Yes. I am not super sure yet about it. Not every major dep bump to the lib is a major bump to consumers. 

Also I need to drop the usage of the action. It doesn't support mono repos.

emmenko: Ah ok, I thought it was a typo or something

PR #1320, opened 2020-11-21
chore: update deps

tdeekens:
#### Summary

Let see.


packages/cypress-plugin/package.json
```diff
@@ -2,11 +2,11 @@
   "name": "@flopflip/cypress-plugin",
   "version": "2.0.0",
   "description": "A plugin for Cypress change feature toggles in Cypress runs",
-  "main": "dist/cypress-plugin.cjs.js",
-  "module": "dist/cypress-plugin.esm.js",
+  "main": "dist/flopflip-cypress-plugin.cjs.js",
+  "module": "dist/flopflip-cypress-plugin.esm.js",
```
tdeekens, line 6: Yay. This a nice change. I wondered if I could convince `preconstruct` in v1 cause in the e.g. the Chrome Dev Tools is can easily become ambiguous without the prefix. /cc @emmenko

PR #1321, opened 2020-11-23
fix: to not use typings on any package

tdeekens:
#### Summary

This seems to cause issues for consumers.


emmenko (commented):
The `types` field also should be removed

tdeekens:
A crap. I think that was causing the issue cause it's outdated.

emmenko:
The issue was because the path didn't match. But preconstruct says that those fields are not necessary, as TS is able to get the types (which I can confirm trying it locally).

PR #1323, opened 2020-11-23
fix: to not infer returns

tdeekens:
#### Summary

This fixes the `(): import('../../types')` issue which seems to be Babel adding them.

### DIFF
--- a/packages/localstorage-adapter/src/adapter/adapter.ts
+++ b/packages/localstorage-adapter/src/adapter/adapter.ts
@@ -70,7 +70,9 @@ const normalizeFlag = (
   // Multi variate flags contain a string or `null` - `false` seems more natural.
   flagValue === null || flagValue === undefined ? false : flagValue,
 ];
-const normalizeFlags = (rawFlags: Readonly<TFlags>) => {
+const normalizeFlags = (
+  rawFlags: Readonly<TFlags>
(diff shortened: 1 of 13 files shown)


tdeekens:
I don't mind typing the returns. As it's public API. If I change them I want TypeScript to yell at me.

packages/localstorage-adapter/src/adapter/adapter.ts
```diff
@@ -70,10 +70,10 @@ const normalizeFlag = (
   // Multi variate flags contain a string or `null` - `false` seems more natural.
   flagValue === null || flagValue === undefined ? false : flagValue,
 ];
-const normalizeFlags = (rawFlags: Readonly<TFlags>) => {
-  if (!rawFlags) return {};
```
emmenko, line 74: Was this line removed on purpose? 🧐

tdeekens: Thanks for checking in. Yes, with an explicit return it is otherwise a type error as it is not a record. The other normalizeFn (each adapter has one) do not have it. I can't recall why I added this.

tdeekens: Actually a test catches me. So I need to refresh my memory a bit ;)

PR #1392, opened 2021-01-13
feat(http-adapter): add new http-adapter

tdeekens:
#### Summary

A more generic http adapter.

### DIFF
--- a/packages/http-adapter/src/adapter/adapter.spec.js
+++ b/packages/http-adapter/src/adapter/adapter.spec.js
@@ -1,0 +1,337 @@
+import { AdapterConfigurationStatus, cacheIdentifiers } from '@flopflip/types';
+import warning from 'tiny-warning';
+import getGlobalThis from 'globalthis';
+import adapter from './adapter';
+
+jest.mock('tiny-warning');
+
+const createAdapterEventHandlers = (custom = {}) => ({
+  onFlagsStateChange: jest.fn(),
+  onStatusStateChange: jest.fn(),
+  ...custom,
+});
+
+describe('when configuring', () => {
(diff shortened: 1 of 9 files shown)


packages/graphql-adapter/src/adapter/adapter.ts
```diff
@@ -8,8 +8,8 @@ import type {
   TFlags,
   TFlagsUpdateFunction,
   TFlagsChange,
-  TGraphQLAdapterInterface,
-  TGraphQLAdapterArgs,
+  TGraphQlAdapterInterface,
+  TGraphQlAdapterArgs,
```
tdeekens, line 12: Unrelated.

tdeekens (commented):
This @emmenko is a very simple `http-adapter`. If you have time please have a look. Start the types. Go to the tests then to the code.

tdeekens:
This has been published from my machine as `0.0.1` so we can try it out already.

packages/http-adapter/src/adapter/adapter.spec.js
```diff
@@ -0,0 +1,337 @@
...
+      it('should invoke and trigger `warning`', () => {
+        expect(warning).toHaveBeenCalledWith(false, expect.any(String));
+      });
+    });
+  });
+});
+
+describe('when configured', () => {
+  let adapterArgs = {
+    execute: jest.fn().mockResolvedValue({ enabled: true, disabled: false }),
```
emmenko, line 44: Should you return a Promise here?

emmenko (commented):
That was quick 😁

Had a look, generally looks good and solid 👌

PR #1403, opened 2021-01-29
feat: add test context provider for react-broadcast

tdeekens:
#### Summary

This feature intends to ease testing setup with the react-broadcast package.

#### Description

Prior to this change you would have to configure flopflip regularily with a memory adapter in a testing setup. 

This has various downsides. One that it puts complexity on setup but also that tests with feature flags become async as flags are flushed through the adapter to state.

With this setup becomes easier and testing too. We can't expose a shared test setup for all packages as a setup with redux requires a store setup. Which itself has a native way of mocking.

### DIFF
--- a/.changeset/modern-nails-approve.md
+++ b/.changeset/modern-nails-approve.md
@@ -1,6 +1,6 @@
 ---
-"@flopflip/react-broadcast": minor
-"@flopflip/react": patch
+'@flopflip/react-broadcast': minor
+'@flopflip/react': patch
 ---
 
 feat: add test context provider for react-broadcast
@@ -8,7 +8,7 @@ feat: add test context provider for react-broadcast
 You can wrap your application using for instance `react-testing-library` as
 
 ```jsx
-import { FlopflipTestProvider } from '@flopflip/react-broadcast';
+import { TestProviderFlopflip } from '@flopflip/react-broadcast';
 import { render } from '@testing-library/react';
 
 const testFlags = {
@@ -18,9 +18,9 @@ const testFlags = {
 describe('rendering', () => {
   it('should render the application with feature flags', () => {
     render(
-      <FlopflipTestProvider flags={testFlags}>
+      <TestProviderFlopflip flags={testFlags}>
         <App />
-      </FlopflipTestProvider>
+      </TestProviderFlopflip>
     );
   });
 });
--- a/.changeset/modern-nails-approve.md
+++ b/.changeset/modern-nails-approve.md
@@ -9,8 +9,19 @@ You can wrap your application using for instance `react-testing-library` as
 
 ```jsx
 import { FlopflipTestProvider } from '@flopflip/react-broadcast';
+import { render } from '@testing-library/react';
 
-<FlopflipTestProvider flags={testFlags} status={adapterStatus}>
-   <App />
-</FlopflipTestProvider>
+const testFlags = {
+  myFlag: true,
+};
+
+describe('rendering', () => {
+  it('should render the application with feature flags', () => {
+    render(
+      <FlopflipTestProvider flags={testFlags}>
+        <App />
+      </FlopflipTestProvider>
+    );
+  });
+});
 ```
--- a/.changeset/modern-nails-approve.md
+++ b/.changeset/modern-nails-approve.md
@@ -1,0 +1,16 @@
+---
+"@flopflip/react-broadcast": minor
+"@flopflip/react": patch
+---
+
+feat: add test context provider for react-broadcast
+
+You can wrap your application using for instance `react-testing-library` as
+
+```jsx
+import { FlopflipTestProvider } from '@flopflip/react-broadcast';
+
+<FlopflipTestProvider flags={testFlags} status={adapterStatus}>
+   <App />
+</FlopflipTestProvider>
+```
--- a/.changeset/modern-nails-approve.md
+++ b/.changeset/modern-nails-approve.md
@@ -8,7 +8,7 @@ feat: add test context provider for react-broadcast
 You can wrap your application using for instance `react-testing-library` as
 
 ```jsx
-import { TestProviderFlopflip } from '@flopflip/react-broadcast';
+import { TestProviderFlopFlip } from '@flopflip/react-broadcast';
 import { render } from '@testing-library/react';
 
 const testFlags = {
@@ -18,9 +18,9 @@ const testFlags = {
 describe('rendering', () => {
   it('should render the application with feature flags', () => {
     render(
-      <TestProviderFlopflip flags={testFlags}>
+      <TestProviderFlopFlip flags={testFlags}>
         <App />
-      </TestProviderFlopflip>
+      </TestProviderFlopFlip>
     );
   });
 });
--- a/packages/react-broadcast/src/components/index.ts
+++ b/packages/react-broadcast/src/components/index.ts
@@ -1,3 +1,4 @@
 export { default as ToggleFeature } from './toggle-feature';
 export { default as ConfigureFlopFlip } from './configure';
 export { default as ReconfigureFlopFlip } from './reconfigure';
+export { default as FlopflipTestContextProvider } from './test-context-provider';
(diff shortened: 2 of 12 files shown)


packages/react-broadcast/src/components/test-context-provider/test-context-provider.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+type TProps = {
+  children: React.ReactNode;
+  flags: TFlags;
+  adapterIdentifiers?: TAdapterIdentifiers & string[];
+  reconfigure?: TReconfigureAdapter;
+  status?: TAdapterStatus;
+};
+
+const defaultProps: Pick<TProps, 'recondigure' | 'status'> = {
```
emmenko:
```suggestion
const defaultProps: Pick<TProps, 'reconfigure' | 'status'> = {
```

packages/react-broadcast/src/components/test-context-provider/test-context-provider.spec.js
```diff
@@ -0,0 +1,73 @@
...
+  it('should expose features', async () => {
+    const rendered = render({
+      flags: {
+        [testFlagName]: true,
+      },
+    });
+
+    await rendered.findByText(/is configured: yes/i);
+
+    expect(rendered.getByText(/feature enabled: yes/i)).toBeInTheDocument();
```
emmenko: Nit: this assertion can be moved inside the `should expose the default adapter status` test, since it's the same setup.

packages/react-broadcast/src/components/test-context-provider/test-context-provider.tsx
```diff
@@ -0,0 +1,52 @@
...
+  adapterIdentifiers: ['test'],
```
emmenko, line 28: What's the `test` adapter?

packages/react-broadcast/src/components/test-context-provider/test-context-provider.tsx
emmenko: Also, shouldn't this be `adapterIdentifiers`? 🤔

```suggestion
const defaultProps: Pick<TProps, 'adapterIdentifiers' | 'status'> = {
```

packages/react-broadcast/src/components/test-context-provider/test-context-provider.tsx
```diff
@@ -0,0 +1,52 @@
...
+  status: {
+    subscriptionStatus: AdapterSubscriptionStatus.Subscribed,
+    configurationStatus: AdapterConfigurationStatus.Configured,
+  }
+};
+
+const TestContextProvider = (props: TProps) => {
```
emmenko, line 35: Can you clarify what makes this setup specific to testing?

packages/react-broadcast/src/components/index.ts
```diff
@@ -1,3 +1,4 @@
 export { default as ToggleFeature } from './toggle-feature';
 export { default as ConfigureFlopFlip } from './configure';
 export { default as ReconfigureFlopFlip } from './reconfigure';
+export { default as FlopflipTestContextProvider } from './test-context-provider';
```
emmenko, line 4: Nit: I think I would just export it without the prefix, people can always alias it when they import it.

```suggestion
export { default as TestContextProvider } from './test-context-provider';
```

packages/react-broadcast/src/components/test-context-provider/test-context-provider.tsx
tdeekens: Yes, thanks. Noticed that too :)

packages/react-broadcast/src/components/test-context-provider/test-context-provider.tsx
tdeekens: Yes. I can't use it internally. The two contexts (flags and adapter) are internally in different parts of the tree. The adpater context higher up in the react package (cause broadcast and redux use it) and the flags context within the respective package.

packages/react-broadcast/src/components/index.ts
tdeekens: Yeah, but I suspect many won't nor know how to. Same with the above. I'd like to make sure things are specifically named for a consumer on the exported components.

.changeset/modern-nails-approve.md
```diff
@@ -0,0 +1,16 @@
...
+feat: add test context provider for react-broadcast
+
+You can wrap your application using for instance `react-testing-library` as
+
+```jsx
+import { FlopflipTestProvider } from '@flopflip/react-broadcast';
+
+<FlopflipTestProvider flags={testFlags} status={adapterStatus}>
+   <App />
+</FlopflipTestProvider>
```
emmenko, line 15:
```suggestion
import { FlopflipTestProvider } from '@flopflip/react-broadcast';
import { render } from '@testing-library/react';

const testFlags = {
  myFlag: true,
};

describe('rendering', () => {
  it('should render the application with feature flags', () => {
    render(
      <FlopflipTestProvider flags={testFlags}>
        <App />
      </FlopflipTestProvider>
    );
  });
});
```

packages/react-broadcast/src/components/index.ts
```diff
@@ -1,3 +1,4 @@
 export { default as ToggleFeature } from './toggle-feature';
 export { default as ConfigureFlopFlip } from './configure';
 export { default as ReconfigureFlopFlip } from './reconfigure';
+export { default as FlopflipTestProvider } from './test-provider';
```
emmenko, line 4: Nit: to keep the name consistent?

```suggestion
export { default as TestProviderFlopFlip } from './test-provider';
```

packages/react-broadcast/src/components/test-provider/test-provider.tsx
```diff
@@ -0,0 +1,58 @@
...
+  children: React.ReactNode;
+  flags: TFlags;
+  adapterIdentifiers?: TAdapterIdentifiers[];
+  reconfigure?: TReconfigureAdapter;
+  status?: TAdapterStatus;
+};
+
+const defaultProps: Pick<
+  TProps,
+  'adapterIdentifiers' | 'reconfigure' | 'status'
```
emmenko, line 26: Is `reconfigure` a default prop?

packages/react-broadcast/src/components/test-context-provider/test-context-provider.tsx
tdeekens: Adapters have an id. This id is used to update only a respective adapter state when you use muliple. As such flag state is `Record<TAdapterIdentifier, TFlags>`. This just make clear that it's test flag state.

.changeset/modern-nails-approve.md
tdeekens: Wow. Thanks!

packages/react-broadcast/src/components/test-context-provider/test-context-provider.tsx
emmenko: Ok so in general other identifiers could be `memory`, `http`, `launchdarkly`, etc. Did I get this right?

So then my next question would be: what would be the use case for the test provider to have any other adapter other than `test`? 🤔

packages/react-broadcast/src/components/index.ts
tdeekens: Sounds odd but consistent :)

.changeset/modern-nails-approve.md
emmenko: Well, since you mentioned `react-testing-library` you should provide an example for that.

You're welcome 😜

packages/react-broadcast/src/components/test-provider/test-provider.tsx
tdeekens: Yes, it defaults to a noop.

packages/react-broadcast/src/components/index.ts
```diff
@@ -1,3 +1,4 @@
 export { default as ToggleFeature } from './toggle-feature';
 export { default as ConfigureFlopFlip } from './configure';
 export { default as ReconfigureFlopFlip } from './reconfigure';
+export { default as TestProviderFlopflip } from './test-provider';
```
emmenko, line 4: Sorry to keep nitpicking this 😅

```suggestion
export { default as TestProviderFlopFlip } from './test-provider';
```

emmenko (approved):

packages/react-broadcast/src/components/index.ts
tdeekens: No thanks. This is good!

PR #1552, opened 2021-11-16
test: react 18

eps1lon:
#### Summary

Re: https://twitter.com/tdeekens/status/1460559774966292482

#### Description

Adjust tests to highlight problems in React 18.

#### Technical debt & future

I had a glance at how the "status" is handled. I've seen some subscription-like patterns so you probably need to use [`useSyncExternalStore`](https://github.com/reactwg/react-18/discussions/86) to avoid the tearing that now occurs in React 18.

Considering test results change depending on how focused tests are run (full suite, single file, single test etc) there's probably also some cleanup issue. If a test schedules a microtask that changes global state then this could affect subsequent test.

### DIFF
--- a/.changeset/chatty-horses-jog.md
+++ b/.changeset/chatty-horses-jog.md
@@ -1,20 +1,20 @@
 ---
-"@flopflip/adapter-utilities": patch
-"@flopflip/combine-adapters": patch
-"@flopflip/cypress-plugin": patch
-"@flopflip/graphql-adapter": patch
-"@flopflip/http-adapter": patch
-"@flopflip/launchdarkly-adapter": patch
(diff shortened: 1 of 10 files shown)


tdeekens:
Thanks for the contribution and investigation!

> I had a glance at how the "status" is handled. I've seen some subscription-like patterns so you probably need to use useSyncExternalStore to avoid the tearing that now occurs in React 18.

Yes, it's basically a subscription which should be treated as an external store. If I recall the adoption strategy correctly, there will be a shim for older version of React so that using the hook won't be a breaking change for the library which is what I'd prefer.

> If a test schedules a microtask that changes global state then this could affect subsequent test.

Good call. Something if I understand right I could look into separately.

PR #1602, opened 2022-03-24
refactor: to use pnpm

tdeekens:
#### Summary

Let's see what we broke.


tdeekens:
Works locally. Can't convince CI.

PR #1691, opened 2022-10-18
fix: to not overwrite internally set adapter args

tdeekens:
#### Summary

Ohterwise internally set adapterArgs through reconfigure are always overwritten by the adapterArgs passed by props.

### DIFF
--- a/packages/react/src/components/configure-adapter/configure-adapter.tsx
+++ b/packages/react/src/components/configure-adapter/configure-adapter.tsx
@@ -410,19 +410,8 @@ const usePendingAdapterArgsEffect = ({
   );
 
   useEffect(() => {
-    /**
-     * NOTE:
-     *   The component might receive `adapterArgs` from `ReconfigureFlopflip`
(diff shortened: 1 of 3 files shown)


packages/react/src/components/configure-adapter/configure-adapter.tsx
```diff
@@ -410,19 +410,8 @@ const usePendingAdapterArgsEffect = ({
...
-     *   passed in will overwrite what `ReconfigureFlopflip` passed in before
-     *   yielding a loss in configuration.
-     *
-     *   However, when the adapter is configured we want the component to
-     *   act in a controlled manner so that overwriting will occur when the
-     *   passed `adapterArgs` change.
-     */
     reconfigureOrQueue(adapterArgs, {
-      shouldOverwrite: getIsAdapterConfigured(),
+      shouldOverwrite: false,
```
tdeekens, line 414: This was a false assumption. Given you reconfigure using the context the component isn't controlled anymore.

NickDevG (approved):
Excellent, amazing how a single property changes that much

PR #1848, opened 2024-02-26
Include new hook to read all enabled feature toggles

CarlosCortizasCT:
#### Summary

Include new hook to read all enabled feature toggles

#### Description

The idea is that there might be use cases where a consumer would like to read all configured feature toggles that are enabled.

In this PR I propose to include a new hook that achieves this feature.

It is not clear to me yet what to hook should return: it currently returns an object of type `TFlags` the same way it's used internally but maybe it would make more sense to just return an array of strings with the names of the enabled feature toggles.

Also, it's not clear to me how internal flags are supposed to use, as there' one configured in the test context and I don't know how the code is supposed to handle those king of toggles.
Are they something actually used? Do they have any specific context to know they should not be exposed?

### DIFF
--- a/packages/react-broadcast/src/hooks/use-all-feature-toggles/index.ts
+++ b/packages/react-broadcast/src/hooks/use-all-feature-toggles/index.ts
@@ -1,0 +1,1 @@
+export { default } from './use-all-feature-toggles';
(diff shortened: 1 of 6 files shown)


CarlosCortizasCT:
The only thing I'm missing is testing how flags are resolved in case there are several adapters configured, but I don't see that tests in the other hooks either and it will require more test-utils updates.

@tdeekens do you think we should aim for it?

tdeekens:
> The only thing I'm missing is testing how flags are resolved in case there are several adapters configured, but I don't see that tests in the other hooks either and it will require more test-utils updates.

Good point. That would require test-utils adjustments indeed to use another memory adapter with another id or using the localstorage-adapter as a second one.

We can leave this out here. The code looks simple enough to me.

tdeekens (approved):

Issue #1879, opened 2024-03-23
A `cachePolicy` to guide behaviour of how cached flags are restored


tdeekens:
A `cachaPolicy` similar to Apollo's `fetchPolicy` could help to steer behaviour of `flopflip` when flags are cachedf in local storage. 

To recap [this](https://www.apollographql.com/docs/react/data/queries/#supported-fetch-policies) is what Apollo defines as `fetchPolicies`.

![CleanShot 2024-03-23 at 22 26 27@2x](https://github.com/tdeekens/flopflip/assets/1877073/198c960d-5fb3-40f7-967d-4b16952ccb53)

Any naming in flopflip can be different and the list does not have to employ the same number.

We have some use cases in flopflip for this.

1. User comes to page, flag is cached.
1.1. Later fetch shows flag has the _same_ value in remote
1.2. Later fetch shows flag has _different_ value in remote
2. User comes to page, new flag is configured
3. User comes to page, flag is not cached
3.1. A default flag is configured
3.2. A default flag is not configured

The above list might not yet be complete but gives an idea for cases.

Currently the `cachePolicies` could be:

1. `stickyCache`

If a cached flag is found, it's not subscribed to. No network request. The cache won't update. A user is "stuck" with the cached flag value until the cache is pruned.

3. `cacheAndNetwork`

No matter if the flag is in the `cache`, a flag subscription is setup and the result is used. The cache will always be updated after the network request.

4. `cacheFirst`

The cache is used if a flag is found in it. A flag subscription is setup and the cache will be updated. The application however is not notified it the flag has a different values form the cache. Only a reload will make that change take effect.

CarlosCortizasCT:
@tdeekens Thanks a lot for this proposal, it looks like a very nice feature for this library to have 💯 

I wanted to also share another point of view if that's ok.

I agree with the proposal to introduce a new attribute which allows users to enable a client-side cache and how to control it, however I'd like to suggest maybe a slightly narrower initial surface for the feature as well as some different naming.

From my point of view, I find it difficult to find a common use case for `stickyCache` as it will need some imperative implementation to clear/update the cached value. Perhaps you already how some use cases in mind. If so, could you share? 🙏 

My idea (at least) for a first implementation is that it will always be a network request (subscription) no matter if the cache is empty or not. The main difference is what we want to do when the remote value is resolved and is different than the cached one. But the process would be the same for a given flag:
1. Read the value from the defaults
2. (if cache is enabled) override the value from the cache
[...]

PR #1888, opened 2024-04-05
feat: to add cache-mode to adapter

tdeekens:
#### Summary

Relates to #1879

This introduces the behaviour of eager and lazy cache modes. At the same time it removes the `subscribeToFlagChanges` as we always subscribe. Lastly, the `unsubscribeFromCachedFlags` is almost like the lazy cache.

### DIFF
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -29,8 +29,8 @@
     "client"
   ],
   "dependencies": {
-    "@flopflip/localstorage-cache": "13.5.2",
-    "@flopflip/sessionstorage-cache": "13.5.2",
-    "@flopflip/types": "13.5.2"
+    "@flopflip/localstorage-cache": "13.6.0",
+    "@flopflip/sessionstorage-cache": "13.6.0",
+    "@flopflip/types": "13.6.0"
   }
 }
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "@flopflip/cache",
+  "version": "13.5.2",
+  "description": "Caching for flipflop adapters",
+  "main": "dist/flopflip-cache.cjs.js",
+  "module": "dist/flopflip-cache.esm.js",
+  "files": [
+    "readme.md",
+    "dist/**"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/tdeekens/flopflip.git",
+    "directory": "packages/cache"
+  },
+  "author": "Tobias Deekens <nerd@tdeekens.name>",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/tdeekens/flopflip/issues"
+  },
+  "homepage": "https://github.com/tdeekens/flopflip#readme",
+  "keywords": [
+    "feature-flags",
+    "feature-toggles",
+    "cache",
+    "client"
+  ],
+  "dependencies": {
+    "@flopflip/localstorage-cache": "13.5.2",
+    "@flopflip/sessionstorage-cache": "13.5.2",
+    "@flopflip/types": "13.5.2"
+  }
+}
--- a/packages/cache/src/cache/cache.ts
+++ b/packages/cache/src/cache/cache.ts
@@ -1,5 +1,4 @@
 import {
-  adapterIdentifiers,
   cacheIdentifiers,
   type TAdapterIdentifiers,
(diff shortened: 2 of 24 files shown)


packages/cache/package.json
```diff
@@ -0,0 +1,36 @@
+{
```
tdeekens, line 1: This makes a lot of things easier. No adapter knows about a direct type (local or session storage) of cache. A facade takes care of delegating and lazily importing concrete caches.

packages/cache/src/cache/cache.ts
```diff
@@ -1,13 +1,16 @@
...
 } from '@flopflip/types';
-import { type LDContext } from 'launchdarkly-js-client-sdk';
 
-const CACHE_PREFIX = '@flopflip/launchdarkly-adapter';
-const FLAGS_KEY = 'flags';
-const FLAGS_REFERENCE_KEY = 'flags-reference';
+const FLAGS_CACHE_KEY = 'flags';
+const FLAGS_REFERENCE_CACHE_KEY = 'flags-reference';
+
+function getCachePrefix(adapterIdentifiers: TAdapterIdentifiers) {
```
tdeekens, line 11: This allows for making it generic.

packages/cache/src/cache/cache.ts
```diff
@@ -41,32 +46,37 @@ async function getCache(
...
   return {
     set(flags: TFlags) {
-      const haveFlagsBeenWritten = flagsCache.set(FLAGS_KEY, flags);
+      const haveFlagsBeenWritten = flagsCache.set(FLAGS_CACHE_KEY, flags);
 
       if (haveFlagsBeenWritten) {
         referenceCache.set(
-          FLAGS_REFERENCE_KEY,
-          [flagsCachePrefix, FLAGS_KEY].join('/')
+          FLAGS_REFERENCE_CACHE_KEY,
```
tdeekens, line 53: Just renaming it.

packages/cache/src/cache/cache.ts
```diff
@@ -41,32 +46,37 @@ async function getCache(
...
+      referenceCache.unset(FLAGS_REFERENCE_CACHE_KEY);
 
-      return flagsCache.unset(FLAGS_KEY);
+      return flagsCache.unset(FLAGS_CACHE_KEY);
     },
   };
 }
 
-function getCachedFlags(cacheIdentifier: TCacheIdentifiers): TFlags {
+function getCachedFlags(
```
tdeekens, line 71: This will later be used in `react-redux` and `react-broadcast`.

packages/graphql-adapter/src/adapter/adapter.ts
```diff
@@ -75,38 +73,6 @@ class GraphQlAdapter implements TGraphQlAdapterInterface {
   readonly #getIsFlagLocked = (flagName: TFlagName) =>
     this.#adapterState.lockedFlags.has(flagName);
 
-  readonly #getCache = async (cacheIdentifier: TCacheIdentifiers) => {
```
tdeekens, line 78: This for each adapter moved to the cache package.

packages/graphql-adapter/src/adapter/adapter.ts
```diff
@@ -149,12 +115,21 @@ class GraphQlAdapter implements TGraphQlAdapterInterface {
...
+              adapterIdentifiers.graphql,
+              this.#adapterState.user?.key
+            );
 
             cache.set(nextFlags);
           }
 
           this.#adapterState.flags = nextFlags;
+
+          if (adapterArgs.cacheMode === 'lazy') {
```
tdeekens, line 129: In lazy mode we just write to the cache. No flushing of updates. The next configuration or reconfiguration (can be a reload) will flush this state "into React".

tdeekens (commented):
Next up I'll have to plug the loading of flags from the cache into `react-redux` and `react-broadcast` _before_ the initial "setState" so that there is no "flash of changing flags".

packages/cache/src/cache/cache.ts
```diff
@@ -98,4 +99,25 @@ function getCachedFlags(
...
-export { getCache, getCachedFlags };
+function getAllCachedFlags(
+  adapter: TAdapter,
+  cacheIdentifier?: TCacheIdentifiers
+) {
+  if (!cacheIdentifier) {
+    return {};
+  }
+
+  if (adapter.effectIds) {
```
tdeekens, line 110: This allows us to load default flags from various adapters when combined.

PR #1922, opened 2024-07-01
refactor: cache keys to be bound to adapter context

tdeekens:
#### Summary

This refactors the cache's key for flags to be a hashed version of the adapter's context.

#### Description

When caching lazily it is important to fall back to a hot cache as often as possible. Otherwise a user would be served a stale cache for flags.

Taking for example the flow of:

1. User logs in
2. Views project A
3. In project A feature is enabled
4. User navigates to project B
5. In project B feature is disabled

With a lazy acting cache, in the above flow the user at 5. would still see an enabled feature. This is rooted in the fact that the cache does not synchronize its state until another hard reload (or unmounting of flopflip). 

The reason for this behaviour is that at any point flopflip's cache is bound - without this change - to a user's key. Any user can only have one set of cached flags. 

This change proposes to take the entire context of the adapter into account for the cache key. This allows a user to have n caches depending on the number of contexts the user is in. This allows a "hot cache" to be restored more likely and flopflip not falling back onto a cold cache often.

### DIFF
--- a/packages/cache/src/cache/cache.spec.js
+++ b/packages/cache/src/cache/cache.spec.js
@@ -1,13 +1,14 @@
 import { adapterIdentifiers, cacheIdentifiers } from '@flopflip/types';
 
 import {
+  encodeCacheContext,
   getAllCachedFlags,
   getCache,
   getCachedFlags,
   getCachePrefix,
 } from './cache';
 
-const cacheKey = 'test';
+const cacheContext = { key: 'some-user-key', value: 'some-other-value' };
 
 describe('general caching', () => {
   it('should allow writing values to the cache', async () => {
@@ -17,7 +18,7 @@ describe('general caching', () => {
     const cache = await getCache(
       cacheIdentifiers.session,
       adapterIdentifiers.memory,
-      cacheKey
+      cacheContext
     );
 
     cache.set(flags);
@@ -32,7 +33,7 @@ describe('general caching', () => {
     const cache = await getCache(
       cacheIdentifiers.session,
       adapterIdentifiers.memory,
-      cacheKey
+      cacheContext
     );
 
     cache.set(flags);
@@ -48,13 +49,15 @@ describe('general caching', () => {
     const cache = await getCache(
       cacheIdentifiers.session,
       adapterIdentifiers.memory,
-      cacheKey
+      cacheContext
     );
 
     cache.set(flags);
 
     expect(sessionStorage.getItem).toHaveBeenLastCalledWith(
-      `${getCachePrefix(adapterIdentifiers.memory)}/${cacheKey}/flags`
+      expect.stringContaining(
+        `${getCachePrefix(adapterIdentifiers.memory)}/${encodeCacheContext(cacheContext)}/flags`
+      )
     );
   });
 });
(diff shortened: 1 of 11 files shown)


packages/cache/src/cache/cache.spec.js
```diff
@@ -1,13 +1,14 @@
...
 import {
+  encodeCacheContext,
   getAllCachedFlags,
   getCache,
   getCachedFlags,
   getCachePrefix,
 } from './cache';
 
-const cacheKey = 'test';
+const cacheContext = { key: 'some-user-key', value: 'some-other-value' };
```
tdeekens, line 11: This is the change of what's passed to the cache now.

packages/cache/src/cache/cache.spec.js
```diff
@@ -78,7 +81,7 @@ describe('flag caching', () => {
       ).toStrictEqual(flags);
 
       expect(sessionStorage.getItem).toHaveBeenLastCalledWith(
-        `${getCachePrefix(adapterIdentifiers.memory)}/${cacheKey}/flags`
+        `${getCachePrefix(adapterIdentifiers.memory)}/${encodeCacheContext(cacheContext)}/flags`
```
tdeekens, line 84: Tests working this way need to get the encoded representation of a cache.

packages/cache/src/cache/cache.ts
```diff
@@ -13,6 +13,20 @@ function getCachePrefix(adapterIdentifiers: TAdapterIdentifiers) {
   return `@flopflip/${adapterIdentifiers}-adapter`;
 }
 
+export function encodeCacheContext(cacheContext: any) {
+  const encodedAsJson = JSON.stringify(cacheContext);
+
+  const hashCode = [...encodedAsJson].reduce(
```
tdeekens, line 19: This is a suggestion at this point. Computing a hashCode might already be enough and is simple. SHA1 also would work but requires - to use native crypto - a site to run under HTTPs. MD5 is another option. Happy to hear feedback.

packages/cache/src/cache/cache.ts
```diff
@@ -34,13 +48,20 @@ async function importCache(cacheIdentifier: TCacheIdentifiers) {
...
+  cacheContext?: any
 ) {
   const cacheModule = await importCache(cacheIdentifier);
 
   const CACHE_PREFIX = getCachePrefix(adapterIdentifiers);
   const createCache = cacheModule.default;
-  const flagsCachePrefix = [CACHE_PREFIX, cacheKey].filter(Boolean).join('/');
+  let encodedCacheContext = '';
+  try {
+    encodedCacheContext = encodeCacheContext(cacheContext);
```
tdeekens, line 59: At times no context may exist of the encoding fails. We then still fall back to caching more globally.

packages/launchdarkly-adapter/src/adapter/adapter.ts
```diff
@@ -155,7 +155,7 @@ class LaunchDarklyAdapter implements TLaunchDarklyAdapterInterface {
         cacheIdentifier,
         adapterIdentifiers.launchdarkly,
         // NOTE: LDContextCommon is part of the type which we never use.
-        this.#adapterState.context?.key as string
+        this.#adapterState.context
```
tdeekens, line 158: We pass this now to the creation of a cache.

CarlosCortizasCT (approved):
This is a great addition 🚀

PR #1923, opened 2024-07-05
refactor: replace eslint with biome

YahiaElTai:
#### Summary

This PR attempts at replacing ESLint with [biome](https://biomejs.dev/)

#### Description
I have added comments below detailing the issues and the current support of biome to make sure things are clear on what is supported and what is not.

### DIFF
--- a/.changeset/empty-keys-perform.md
+++ b/.changeset/empty-keys-perform.md
@@ -1,0 +1,7 @@
+---
+"@flopflip/adapter-utilities": patch
+"@flopflip/cache": patch
+"@flopflip/combine-adapters": patch
+---
+
+Replace eslint with biome


.github/workflows/test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         run: pnpm typecheck
 
       - name: Lint
-        run: pnpm lint
+        run: pnpm lint:ci
```
YahiaElTai, line 39: biome CLI has a command called `biome ci` which runs the formatter, linter and the import sorter in read-only mode meaning, it won't change any files at all which is what I used here for CI

biome.json
```diff
@@ -0,0 +1,97 @@
+{
+  "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
+  "vcs": {
```
YahiaElTai, line 3: biome allows integrating with `git` to read `gitignore` content and processing only staged files. these features are opted in and not enabled by default. if needed they can be enabled.

As of now I have enabled only reading the ignore file with `useIgnoreFile`

biome.json
```diff
@@ -0,0 +1,97 @@
...
+    "clientKind": "git",
+    "enabled": true,
+    "useIgnoreFile": true
+  },
+  "files": {
+    "ignoreUnknown": true
```
YahiaElTai, line 9: biome currently doesn't support a lot of extensions. You can see the list here https://biomejs.dev/internals/language-support/

This tells biome to ignore them instead of throwing an error

biome.json
```diff
@@ -0,0 +1,97 @@
...
+  },
+  "formatter": {
+    "enabled": true,
+    "formatWithErrors": false,
+    "indentStyle": "space",
+    "indentWidth": 2,
+    "lineEnding": "lf",
+    "lineWidth": 80,
```
YahiaElTai, line 17: biome doesn't have support for `.editorconfig` yet so these values can be added here like lineEnding and others, however this feature is coming soon. Here's [the issue](https://github.com/biomejs/biome/issues/1724).

Keep in mind that biome uses a lot of feature flag so most of the issues that were recently merged are hidden or not even released. so they will come slowly

biome.json
```diff
@@ -0,0 +1,97 @@
...
+    "ignore": [
+      "**/package.json",
+      "**/dist/**",
+      "**/.changeset",
+      "**/CHANGELOG.md",
+      "**/pnpm-lock.yaml",
+      "**/pnpm-workspace.yaml"
+    ]
+  },
+  "organizeImports": { "enabled": true },
```
YahiaElTai, line 27: the import sorter is quite interesting, it sorts import by the closet it you from the current file which is quite helpful once you get used to it.
Read more here https://biomejs.dev/analyzer/import-sorting/

biome.json
```diff
@@ -0,0 +1,97 @@
...
+  "linter": {
+    "enabled": true,
+    "rules": {
+      "recommended": true,
```
YahiaElTai, line 31: I migrated the eslint file which added a lot of rules so I compared these rules to the list of recommended rules by biome and then removed all the redundant and decided to enable all the recommended rules.

This ensures that we follow the recommended rules and have a clean config file for biome.

find the recommended rules here https://biomejs.dev/linter/rules/#recommended-rules

biome.json
```diff
@@ -0,0 +1,97 @@
...
+      "it",
+      "describe",
+      "expect",
+      "jest",
+      "before",
+      "beforeAll",
+      "beforeEach",
+      "after",
+      "afterAll",
+      "afterEach"
```
YahiaElTai, line 94: currently biome doesn't have the idea or `env` or `project`, so it can't really recognize any globals beyond the window object or Node globals so listing all the globals here is currently the only way to do it.

Of course there's currently a [discussion](https://github.com/biomejs/biome/discussions/689) around it and is in progress, I think it should be available in the next major version V2

lint-staged.config.js
```diff
@@ -1,8 +1,6 @@
 module.exports = {
-  'demo/**/*.js': ['pnpm run fix:eslint', 'pnpm run format:ts', 'git add -u'],
-  'packages/**/*.{js,ts}': [
-    'pnpm run fix:eslint',
-    'pnpm run format:ts',
+  '*.{js,ts,tsx}': [
+    'biome check --write --no-errors-on-unmatched', // Format, sort imports, lint, and apply safe fixes
```
YahiaElTai, line 3: `biome check` is another CLI command that is recommended to use in lint-staged, it checks and you can optionally modify files without committing.

The check command is nice because it shows the idea behind biome which is to act as a single tool and not separate tools. you can format, sort imports and lint with a single command. I like it

YahiaElTai: also `tsx` was not included so I added it 😄

package.json
```diff
@@ -7,14 +7,16 @@
...
-    "develop": "jest --projects jest.*.config.js --watch",
-    "lint": "jest --config jest.lint.config.js --maxWorkers 5",
+    "develop": "jest --projects jest.test.config.js --watch",
+    "lint": "biome lint",
     "typecheck": "tsc --project tsconfig.json --noEmit && tsc --project packages/cypress-plugin/tsconfig.json --noEmit",
     "format": "pnpm format:ts && pnpm format:md && pnpm format:yaml",
-    "format:ts": "prettier --write '**/*.{js, ts, tsx}'",
+    "format:ts": "biome format --write",
     "format:md": "prettier --parser markdown --write '**/*.md'",
     "format:yaml": "prettier --parser yaml --write '**/*.{yml,yaml}'",
```
YahiaElTai, line 16: As mentioned above biome language support is quite limited so yaml and md files are not supported yet which is why I kept prettier for that.

package.json
```diff
@@ -70,26 +72,12 @@
...
     "@types/node": "20.14.9",
-    "@typescript-eslint/eslint-plugin": "7.14.1",
-    "@typescript-eslint/parser": "7.14.1",
     "babel-jest": "29.7.0",
     "babel-plugin-transform-dynamic-import": "2.1.0",
     "check-node-version": "4.2.1",
     "colors": "1.4.0",
     "core-js": "3.37.1",
     "cross-env": "7.0.3",
-    "eslint": "8.57.0",
```
YahiaElTai, line 81: This feels right.

packages/adapter-utilities/src/expose-globally/expose-globally.spec.js
```diff
@@ -10,7 +10,7 @@ const testAdapter = {
 it('should expose `adapter` globally', () => {
   exposeGlobally(testAdapter);
 
-  const globalThis = getGlobalThis();
+  const global = getGlobalThis();
```
YahiaElTai, line 13: biome doesn't like shadowing global variables

packages/cache/src/cache/cache.spec.js
```diff
@@ -94,11 +94,6 @@ describe('flag caching', () => {
         flag2: true,
       };
 
-      const memoryAdapterCache = await getCache(
```
YahiaElTai, line 97: was not used so deleted it

packages/combine-adapters/src/adapter/adapter.ts
```diff
@@ -78,9 +77,9 @@ class CombineAdapters implements TCombinedAdapterInterface {
       return;
     }
 
-    this.#adapters.forEach((adapter) => {
+    for (const adapter of this.#adapters) {
       adapter.updateFlags(flags, options);
-    });
+    }
```
YahiaElTai, line 82: biome not a big fan of `forEach`

packages/graphql-adapter/src/adapter/adapter.ts
```diff
@@ -165,18 +168,20 @@ class GraphQlAdapter implements TGraphQlAdapterInterface {
...
+        if (this.#getIsFlagLocked(normalizedFlagName)) {
+          return updatedFlags;
+        }
 
         if (options?.lockFlags) {
           this.#adapterState.lockedFlags.add(normalizedFlagName);
         }
 
-        updatedFlags = {
+        const updated = {
```
YahiaElTai, line 179: this was updating a function argument, biome didn't like that

packages/launchdarkly-adapter/src/adapter/adapter.spec.js
```diff
@@ -405,6 +404,7 @@ describe('when configuring', () => {
             expect(onFlagsStateChange).toHaveBeenCalledTimes(1);
           });
 
+          // biome-ignore lint/complexity/noExcessiveNestedTestSuites: these test suits could be reorganized
```
YahiaElTai, line 407: biome has a limit of 5 nesting levels for describe. I think it's a good rule of thumb to follow, maybe we could reorganize these tests later on

packages/react-broadcast/src/components/test-provider/test-provider.tsx
```diff
@@ -1,16 +1,17 @@
...
-  adapterIdentifiers,
   AdapterSubscriptionStatus,
   type TAdapterIdentifiers,
   type TAdaptersStatus,
   type TFlags,
   type TReconfigureAdapter,
+  adapterIdentifiers,
 } from '@flopflip/types';
+// biome-ignore lint/style/useImportType: <explanation>
 import React from 'react';
```
YahiaElTai, line 12: React is used as a type and sometimes as a Component which confuses both TS and Biome. disabled it

packages/react-redux/src/hooks/use-feature-toggles/use-feature-toggles.ts
```diff
@@ -21,7 +17,9 @@ export default function useFeatureToggles(flags: TFlags) {
         flagVariation
       );
 
-      return [...previousFlags, isFeatureEnabled];
```
YahiaElTai, line 24: biome thinks that the spread operator is evil.

packages/react/src/components/configure-adapter/configure-adapter.tsx
```diff
@@ -58,6 +57,7 @@ const useAppliedAdapterArgsState = ({
   const [appliedAdapterArgs, setAppliedAdapterArgs] =
     useState<TAdapterArgs>(initialAdapterArgs);
 
+  // biome-ignore lint/correctness/useExhaustiveDependencies: <explanation>
```
YahiaElTai, line 60: I didn't want to change the deps here because that could change the behavior and I didn't want to risk it now

packages/combine-adapters/src/adapter/adapter.ts
```diff
@@ -111,7 +110,9 @@ class CombineAdapters implements TCombinedAdapterInterface {
     }
 
     const handleStatusChange = (nextStatus: TAdapterStatusChange['status']) => {
-      if (this.#getIsAdapterUnsubscribed()) return;
+      if (this.#getIsAdapterUnsubscribed()) {
+        return;
+      }
```
YahiaElTai, line 115: This repeats a lot, biome doesn't like return without curely braces (readability)

lint-staged.config.js
YahiaElTai: I updated this to include all files  with `*` hope that's okay. it's easier to maintain and allow for biome to evolve without manual updates

tdeekens (approved):
Great. Thanks for your first contribution.

PR #2183, opened 2026-08-17
fix: start LaunchDarkly client before waiting for initialization

nima-ct:
#### Summary

Start the LaunchDarkly client before waiting for initialization.

#### Description

LaunchDarkly JavaScript SDK v4 separates client creation from startup. createClient() no longer starts initialization automatically, so the adapter timed out while waiting and continued with default flag values.

This change calls client.start() before waitForInitialization() and adds regression coverage for the required call order.

### DIFF
--- a/.changeset/bright-flags-start.md
+++ b/.changeset/bright-flags-start.md
@@ -1,0 +1,5 @@
+---
+"@flopflip/launchdarkly-adapter": patch
+---
+
+fix: start the LaunchDarkly client before waiting for initialization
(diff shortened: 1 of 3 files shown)


tdeekens (approved):
Thanks. Missed that in the upgrade. A contribution - yay 🫂.

tdeekens:
@nima-ct released in https://github.com/tdeekens/flopflip/releases/tag/v15.1.12

Had a bit of pain with oudated setups and updates in the release process but not all is well.

